PR #8, opened 2020-05-14
Fibbonaci

Vicx95:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,25 @@
...
+        y = result;
+    }
+    return result;
 }
 
 int fibonacci_recursive(int sequence) {
     // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0)return 0;
+    if(sequence < 3) return 1;
```
Kasperice, line 23: Second Fibonacci element will be calculated in recursion below. It is not necessary to use this conditional, you can try using  ```if (sequence == 0)``` and ```if (sequence == 1)```. It can be even done under one ```if```.

Vicx95: You are right :)

ziobron:
rid:1119 - punktualność +2XP

ziobron:
rid:1430 - fibonacci +6XP

PR #9, opened 2020-05-14
Impelentation completed

PierwszyKrytykRP:


module1/homework/calculate/calculate.hpp
```diff
@@ -3,5 +3,24 @@
...
+
+    if("add" == command)
+        result = first + second;
+    else if("subtract" == command)
+        result = first - second;
+    else if("multiply" == command)
+        result = first * second;
+    else if("divide" == command)
+    {
+        if(0 == second)
```
Kasperice, line 17: I think ```if (second == 0)``` will be more clear, but as you can see ```if (0 == second)``` also works

Kasperice: It also applies for all of your conditionals

PierwszyKrytykRP: I consider `"divide" == command` to be safer. I won't accidentally assign value of right side to the left side of comparison if i forget one `=`. I would get compile error

ziobron:
rid:1429 - calculate +5XP

ziobron:
rid:1119 - punktualność +2XP

PR #15, opened 2020-05-14
Homework fibbonacci first version

MonikaZelechowska:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,27 @@
...
+    return result;
 }
 
 int fibonacci_recursive(int sequence) {
+
     // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0)
+        return 0;
+    if(sequence == 1 || sequence ==2)
```
Kasperice, line 25: It is not necessary to return 1 when sequence == 2, it will be calculated in recursion

ziobron:
{
"requirement_ids": [1119, 1430],
"points": [2, 6]
}

PR #28, opened 2020-05-14
complete calculator

cnuebred:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,34 @@
...
 #include <string>
+#include <algorithm>
+#include <cstdlib>
+
+const std::string commands[5] = {"add", "subtract", "multiply", "divide", "quit"};
 
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+    auto x = std::find(begin(commands), end(commands), command);
```
ziobron, line 9: what can happen when the given value is not present in the array? :)
It's not a safe solution :)

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,34 @@
...
+    if(*x == command){
+        auto first_letter = command[0];
+        switch (first_letter)
+        {
+        case 'a':
+            return std::to_string(first + second); 
+            break;
```
ziobron, line 16: you don't need `break` after `return`. Function execution will be finished anyway.

ziobron:

PR #32, opened 2020-05-15
Calculate

Kasperice:


module1/homework/calculate/calculate.hpp
```diff
@@ -3,5 +3,20 @@
...
+        if (second != 0) {
+            return std::to_string(first / second);
+        } else {
+            return "Division by 0";
+        }
+    } else {
+        return "Invalid data";
+    }
     return "";
 }
```
Marcinati, line 22: Previous return will cover rest, I don't think this is nessesary.

Kasperice: Thanks, I missed that

ziobron:

PR #34, opened 2020-05-15
All operations work

kozlofski:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,33 @@
...
+    else if (command == "multiply") {
+        std::string result = std::to_string(first * second);
+        return result;
+    }
+    else if (command == "divide") {
+        std::string result;
+        if (second == 0) {
+            result = "Division by 0";
+        }
+        else {
```
kuchichan, line 23: Możesz pominąć ten else jeśli zwrócisz od razu "Division by 0"

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,33 @@
...
 #include <string>
 
-std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+std::string calculate(const std::string &command, int first, int second)
+{
+    if (command == "add") {
+        std::string result = std::to_string(first + second);
+        return result;
```
DamiXioR, line 8: in my opinion you used too many variables, it's enough to use return + action

kozlofski: Yes, I know :) I had an error here, I was unable to return it in one line, so I added this variable in those all lines, but the error was somewhere else and I didn't simplify it after.

ziobron:

PR #37, opened 2020-05-15
Homework push fib + calc

kszytko:
Testing merging

### DIFF
--- a/module1/homework/fibonacci/fibonacci.hpp
+++ b/module1/homework/fibonacci/fibonacci.hpp
@@ -1,11 +1,25 @@
 #pragma once
+#include <vector>
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    std::vector<int> fib(sequence + 1);
+
+    fib[0] = 0;
+    fib[1] = 1;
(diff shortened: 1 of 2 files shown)


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,25 @@
...
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    std::vector<int> fib(sequence + 1);
+
+    fib[0] = 0;
+    fib[1] = 1;
+
+    for (int i = 2; i < sequence + 1; i++){
+        fib[i] = fib[i-1] + fib[i-2];
```
ziobron, line 11: a clever way to get the simplicity of recursive algorithm. It has a price of allocation of every element from the beginning, so it works in the same way. For small numbers, it's not a problem, but it will require a lot of memory for large numbers.

ziobron:
{ "requirement_ids": [1119, 1429, 1430], "points": [2, 5, 6] }

ziobron:
{ "requirement_ids": [1119], "points": [2] }

PR #39, opened 2020-05-15
Fibonacci

kari0036:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,26 @@
...
+
+    }
+    return score;
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if(sequence==0)return 0;
+    if(sequence <= 2)return 1;
```
ziobron, line 23: inconsistent formatting. Please apply spaces before and after operators like ==, +, -
```suggestion
    if (sequence == 0) return 0;
    if (sequence <= 2) return 1;
```

ziobron:

PR #53, opened 2020-05-16
fib rec, fib ite works

Marcinati:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,40 @@
 #pragma once
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
+
+    long int fibi[23601];
```
Kasperice, line 5: Fibonacci_iterative solution can be done on 3 variables, there is no need to allocate such a big array

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,40 @@
...
+
+    }
     return 0;
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
+    if (sequence == 0) {
+        return 0;
+    } else if (sequence == 1 || sequence == 2) {
```
Kasperice, line 34: Second condition ```sequence == 2``` is not necessary. It will be covered in ```else``` part, it will take fibonacci_recursive(1) and fibonacci_recursive(0)

Marcinati: True.

module1/homework/fibonacci/fibonacci.hpp
Marcinati: Especially when the declaration of function is 'int' type it's no point to use 'long int' in definition (in this case), because in the end it will return 'int' (correct or not correct when it will go out of range).

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,37 @@
...
+            }
+            return fibi[sequence];
+        }
+    }
     return 0;
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
+    if ((sequence == 0) || (sequence == 1)) {
```
Kasperice, line 31: This can be even shorter

Marcinati:
> Array not necessary. Try simplify solution.
> 
> +6 XP fibonacci
> +2 XP before deadline

Done, using 3 integers.

ziobron:

PR #57, opened 2020-05-16
Calculate

kuchichan:
I'm trying to avoid typical else if ladder and study some functional possibilities in C++

### DIFF
--- a/module1/homework/calculate/calculate.hpp
+++ b/module1/homework/calculate/calculate.hpp
@@ -1,7 +1,24 @@
 #pragma once
 #include <string>
+#include <map>
+#include <functional>
 
-std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+
+std::string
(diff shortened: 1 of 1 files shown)


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,21 @@
...
-std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+std::string calculate(const std::string &command, int first, int second) {
+  
+  std::map<std::string, std::function<int(int,int)>> mapOperator;
+  mapOperator["add"] = std::plus<int>();
+  mapOperator["subtract"] = std::minus<int>();
+  mapOperator["divide"] = std::divides<int>();
+  mapOperator["multiply"] = std::multiplies<int>();
```
ziobron, line 12: It's better to initialize the map with those values in a constructor rather than adding each of them separately.

ziobron:

PR #59, opened 2020-05-16
mikailromer: Define calculate function.

mikailromer:
Merge changes from mikailromer/kurs_cpp_podstawowy repository. Define calculate function, which is able to add, subtract, multiply and divide the given integer numbers.


ziobron:

PR #66, opened 2020-05-17
Implement fibonacci functions

mikailromer:
Define functions counting the Fibonacci number recursively and iteratively.


ziobron:

PR #72, opened 2020-05-17
Fibo working

kuchichan:
Quite "regular" version.

### DIFF
--- a/module1/homework/fibonacci/fibonacci.hpp
+++ b/module1/homework/fibonacci/fibonacci.hpp
@@ -1,11 +1,21 @@
 #pragma once
+#include <cstddef>
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    int a = 0;
+    int b = 1;
+    int temp;
+    for (std::size_t i = 0; i < sequence; ++i) {
+      temp = b;
+      b = a + b;
+      a = temp;
+    }
+    return a;
 }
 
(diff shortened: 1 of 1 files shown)


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,21 @@
 #pragma once
+#include <cstddef>
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    int a = 0;
+    int b = 1;
+    int temp;
+    for (std::size_t i = 0; i < sequence; ++i) {
```
TalNovatzky, line 8: why do you need it std::size_t here instead of an integer?

kuchichan: TBH No serious reason :) I thought that size_t is more adequate during iteration than simple int
(from cpp reference)

> std::size_t is commonly used for array indexing and loop counting. Programs that use other types, such as unsigned int, for array indexing may fail on, e.g. 64-bit systems when the index exceeds UINT_MAX or if it relies on 32-bit modular arithmetic. 

Is that wrong in some way?

TalNovatzky: no, I was just curious, thank you for the answer.

TalNovatzky: now you learned me something ;)

ziobron:

PR #74, opened 2020-05-17
adding works

McKucia:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,17 @@
...
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+    if(command == "add")
+        return std::to_string(first + second);
+    if(command == "subtract")
+        return std::to_string(first - second);
+    if(command == "divide")
+    {
+    	if(second==0 || first ==0)
```
Vicx95, line 10: unnecessary checking that first == 0, because for example you can divide 0 / 5, the result is 0

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,17 @@
...
+    		return "Division by 0";
+        return std::to_string((int)((double)first/(double)second));
```
Vicx95, line 12: why did you convert int variables for double and finnaly you convert result to int? :)

McKucia: hah, I don't know to be honest. I think i was trying to do smth

module1/homework/calculate/calculate.hpp
McKucia: sure

McKucia: sure

PR #75, opened 2020-05-17
add calculate function

kontrybutor:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,28 @@
...
+  });
+  calculator.try_emplace("subtract", [&](const auto first, const auto second) {
+    return std::to_string(first - second);
+  });
+  calculator.try_emplace("multiply", [&](const auto first, const auto second) {
+    return std::to_string(first * second);
+  });
+  calculator.try_emplace("divide", [&](const auto first, const auto second) {
+    return second != 0 ? std::to_string(first / second) : "Division by 0";
+  });
```
ziobron, line 21: It's better to initialize the map with these values rather than put them in separate instructions.
Good idea with map :)

ziobron:

PR #76, opened 2020-05-17
add fibonacci functions

kontrybutor:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,23 @@
...
+
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
+
+  if (sequence == 0) {
     return 0;
+  }
+  std::vector<int> v(sequence, 1);
+  std::adjacent_difference(v.begin(), v.end() - 1, v.begin() + 1,
+                           std::plus<int>{});
```
MechaniQ1, line 14: Nice idea, but you can also do that with 3 variables only. Less memory needed.

kontrybutor: Thanks, fully agree. I measured execution time and solution with for-loop and 3 variables is approx. 10 times faster than my solution.  Anyway, I would like to keep my solution as I promised myself to get to know as much of STL algorithms as possible.

nauka-programowania-MA: using STL is safe, because STL is already tested :)

ziobron:

PR #78, opened 2020-05-17
Fibonacci

DamiXioR:
Fibonacci by Damixior.

### DIFF
--- a/module1/homework/fibonacci/fibonacci.hpp
+++ b/module1/homework/fibonacci/fibonacci.hpp
@@ -1,11 +1,20 @@
 #pragma once
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
+
+
     return 0;
+
 }
 
+
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+
+     if (sequence <= 1) {
+
+     return sequence;
+
+     }
+
+     return fibonacci_recursive(sequence-1) + fibonacci_recursive(sequence-2);
 }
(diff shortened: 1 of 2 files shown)


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,9 +2,33 @@
...
+
+    int temp = tab[1];
+    tab[1] = tab[0] + tab[1];
+    tab[0] = temp;
+
+    sequence--;
+
+    }
+
+    return tab[1];
```
kontrybutor, line 29: It seems you have an error in this line :)

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,9 +2,33 @@
 
 int fibonacci_iterative(int sequence) {
 
+    int tab [] = {0,1};
+
+    if (sequence == 0) {
```
kontrybutor, line 7: You can simplify this few lines into:
```cpp
if (sequence <= 1)
{
    return sequence;
}

module1/homework/fibonacci/fibonacci.hpp
DamiXioR: what king of error? Everything works

kontrybutor: Okay. My bad. You’re right.

module1/homework/fibonacci/fibonacci.hpp
DamiXioR: :)

ziobron:

PR #83, opened 2020-05-17
Homework

TalNovatzky:


ziobron:
Hmm... it looks that you have done homework from Podstawy #2 in the same branch as Podstawy #1. It should be delivered on another branch. Is there also another PR? If so, can you link here to it with #PR_number?

ziobron:
Points for homework 2 awarded in #223. Let me close this PR.

PR #85, opened 2020-05-17
Homework - calculate

DawidMalarz:


TalNovatzky:
You can always check with files are checked out by : **git status** command

PR #88, opened 2020-05-17
Both fibonacci works

malinaq3:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,31 @@
 #pragma once
+#include<iostream>
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+
+
+int tab[100];
```
kontrybutor, line 7: I think you can change it to ```int tab [sequence]``` or even use a ```std::array``` :)

TalNovatzky, line 7: why do you need to use an array here?
you are allocating too much memory here - you can try to do it only with integers and loop.

malinaq3: I changed it to solution with only integers and loop. Thanks!

malinaq3: I don' know how to do it with `std::array`

ziobron:

PR #91, opened 2020-05-18
adding fibonacci

MechaniQ1:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,23 @@
...
+        return fib2;
+    }
+    
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if (sequence == 0) return 0;
+    if (sequence == 1) return 1;
```
jakjan95, line 21: You can use same condition as you did in iterative version.

MechaniQ1: You are right, I will change it later. Thanks for your advice :D

ziobron:

PR #94, opened 2020-05-18
Calculate

piezol:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,17 @@
...
+    if(command == "add") {
+        return std::to_string(first + second);
+    } else if(command == "subtract") {
+        return std::to_string(first - second);
+    } else if (command == "multiply") {
+        return std::to_string(first * second);
+    } else if (command == "divide") {
+        if(second == 0)
+            return "Division by 0";
+        return std::to_string(first / second);
```
jakjan95, line 14: You can cut in to one line by using question mark notation:
_conditional Expression ? expression1 : expression2;_

piezol: Should i, though?
With a separate 'if' statement edge case stands out instead of being mangled into main logic.

Would leave that as it is (opinion based).

nauka-programowania-MA: option with multiple if else is better

ziobron:

PR #95, opened 2020-05-18
Fibonacci

piezol:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,19 @@
...
+        fib_n = temporary_sum;
+    }
+    return fib_n;
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if (sequence == 0) return 0;
+    if (sequence == 1) return 1;
```
jakjan95, line 17: You can simplify it by making just one condition : sequence <=1 and inside it returning a sequence.

piezol: I would not call it 'simplify', as fibonacci sequence is explicitly defined as (F0 = 0, F1 = 1), and it 'just so happens' that it is logically equivalent to (Fn = n for every non-negative integer <= 1), but it's not a natural way humans look at the sequence.

(i assume we don't go to asm and compare which one is better - yes, yours is usually better, but performance was not mentioned).

nauka-programowania-MA: I agree with jakjan95 that sequence <=1 is better option.

ziobron:

PR #96, opened 2020-05-18
Add calculate function

mateusz-talma:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,34 @@
...
-    return "";
+
+    std::string res;
+    if (!command.compare(operations::add))
+        res = std::to_string(first + second);
+    else if (!command.compare(operations::sub))
+        res = std::to_string(first - second);
+    else if (!command.compare(operations::mul))
+        res = std::to_string(first * second);
+    else if (!command.compare(operations::div))
```
ziobron, line 22: we usually don't use `compare` function on strings. You can just use ==. It has the same effect and is more readable.

ziobron:

PR #97, opened 2020-05-18
Add Fibonacci functions

mateusz-talma:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,36 @@
...
-    // TODO: Your implementation goes here
-    return 0;
+int fibonacci_recursive(int sequence) 
+{
+    if (sequence <= 0)
+        return 0;
+    else if (sequence <= 2)
+        return 1;
+    else
+        return fibonacci_recursive(sequence - 1) + fibonacci_recursive(sequence - 2);
```
jakjan95, line 35: Good practice is to use {} in if statement, in case of conditions you can make it simpler just by checking it once
```suggestion
    if (sequence <= 1){
        return sequence;
    }
    return fibonacci_recursive(sequence - 1) + fibonacci_recursive(sequence - 2);
```
In test we don't have scenario for negative numbers, but it's good to return error code or exception for negative numbers

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,36 @@
...
+        else if (i == 1)
+        {
+            prevFib.at(1) = 1;
+            continue;
+        }
+        fib = prevFib.at(0) + prevFib.at(1);
+        prevFib.at(0) = prevFib.at(1);
+        prevFib.at(1) = fib;
+    }
+    return fib;
```
jakjan95, line 25: In case of conditions you can use same as in recrusive version.
In case of iterative version you can make it simpler by just using 3 variables for 2 previous fibonacci numbers and for present fibonacci number.

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,34 @@
...
-    return "";
+
+    std::string res;
+    if (!command.compare(operations::add))
+        res = std::to_string(first + second);
+    else if (!command.compare(operations::sub))
+        res = std::to_string(first - second);
+    else if (!command.compare(operations::mul))
+        res = std::to_string(first * second);
+    else if (!command.compare(operations::div))
```
jakjan95, line 22: You can simplify it by passing the command directly to the condition.

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,34 @@
...
+        {
+            if (second == 0)
+                res = operations::divByZero;
+            else
+                res = std::to_string(first / second);
+        }
+    else
+        res = "Invalid data";
+
```
jakjan95, line 31: For last condition you can simplify it to one line by using question mark notation.

ziobron:

Points for calculate awarded in #96

PR #98, opened 2020-05-18
+ updated calculate.hpp - working function calculate

MarekJanek:


ziobron:

PR #100, opened 2020-05-18
updated fibonacci.hpp file, fibonacci_recursive and fobocnacci_iterat…

MarekJanek:
…ive working

### DIFF
--- a/module1/homework/fibonacci/fibonacci.hpp
+++ b/module1/homework/fibonacci/fibonacci.hpp
@@ -1,11 +1,31 @@
 #pragma once
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    
+    if(sequence==0)
+        return 0;
+    if(sequence<=2)
+        return 1;
+
+    int res=1, first=0, second=1;
+
+    for(int i = 0;i<sequence-2;i++){
+
+        first=res;
+        res +=second;
+        second = first;
+
+    }
+    return res;
+    
 }
 
 int fibonacci_recursive(int sequence) {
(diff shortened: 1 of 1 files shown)


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,31 @@
...
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+
+    if(sequence==0)
+        return 0;
+    if(sequence<=2)
+        return 1;
```
DamiXioR, line 28: the same effect as if(sequence <= 1) return sequence;

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,31 @@
 #pragma once
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    
+    if(sequence==0)
+        return 0;
+    if(sequence<=2)
+        return 1;
```
DamiXioR, line 8: the same effect as if(sequence <= 1) return sequence;

WojtekMs: I'm wondering if it's not better to leave it the way it is - I think it's easier to understand.

Marcinati: I think:

if(sequence <= 1) {
  return sequence;
}

this is easy to understand the same as your original code, maybe even easier, because you are returning the same sentence which you compare.

ziobron:

PR #101, opened 2020-05-18
Calculate

g00fer:
Nick: Bartek8

### DIFF
--- a/module1/homework/calculate/calculate.hpp
+++ b/module1/homework/calculate/calculate.hpp
@@ -1,7 +1,31 @@
 #pragma once
 #include <string>
 
+
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+
+    if ("add" == command)
+    {
+        return std::to_string( first + second);
+    }
+    else if ("multiply" == command)
+    {
+        return std::to_string( first * second);
+    }
+    else if("subtract" == command)
+    {
+        return std::to_string( first - second);
+    }
+    else if ("divide" == command)
+    {
+        if( 0 != second)
+        {
+            return std::to_string( first / second);
+        }
+        else
+        {
+            return "Division by 0";
+        }
+    }   
+    return "Invalid data";
 }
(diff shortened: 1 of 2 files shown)


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,24 @@
 
 int fibonacci_iterative(int sequence) {
     // TODO: Your implementation goes here
-    return 0;
+    int fib1 = 0;
+    int fib2 = 1;
+    int retval;
+
+    if(0 == sequence) return 0;
+    if(1 == sequence) return 1;
```
DamiXioR, line 10: shorter solution if(sequence <= 1) return sequence;

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,24 @@
...
+        fib2 = retval;
+    }
+    return retval;
 }
 
 int fibonacci_recursive(int sequence) {
     // TODO: Your implementation goes here
-    return 0;
+    if(0 == sequence) return 0;
+    if(1 == sequence) return 1;
```
DamiXioR, line 23: the same if(sequence <= 1) return sequence;

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,31 @@
...
+    else if ("divide" == command)
+    {
+        if( 0 != second)
+        {
+            return std::to_string( first / second);
+        }
+        else
+        {
+            return "Division by 0";
+        }
```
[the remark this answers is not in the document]
anraf1001: The most probable / default option is `second != 0` (or `0 != second`) so in my opinion it's more clear. Anyway I agree that g00fer can delete `else`.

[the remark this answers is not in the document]
g00fer: I think it is better to have explicit reason for 'return' if multiple 'returns' are used in function. Typically it is better to have only single 'return' statement in a function.

module1/homework/fibonacci/fibonacci.hpp
g00fer: I translated polish wikipedia directly into software: "Pierwszy wyraz jest równy 0, drugi jest równy 1, każdy następny jest sumą dwóch poprzednich."

module1/homework/fibonacci/fibonacci.hpp
g00fer: Thank you for the review!

ziobron:

PR #103, opened 2020-05-19
Exercise 2: Fibonacci

kawapa:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,31 @@
...
-    // TODO: Your implementation goes here
-    return 0;
+    int first = 0;
+    int second = 1;
+    int temp;
+
+    if (sequence == 0)
+        return first;
+    if (sequence < 2)
+        return second;
```
Marcinati, line 11: You can return sequence if sequence is less than 2.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,31 @@
...
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
-}
+    if (sequence == 0)
+        return 0;
+    if (sequence < 3)
+        return 1;
```
Marcinati, line 27: Same here, you can return sequence is sequence is less than 2.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,31 @@
...
+
+    return fibonacci_recursive(sequence - 2) +
+        fibonacci_recursive(sequence - 1);
```
Marcinati, line 30: This will work also for sequence equals to 2.
fibonacci_recursive(0) + fibonacci_recursive(1) = 1

ziobron:

PR #104, opened 2020-05-19
Calculate.hpp

WojtekMs:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,28 @@
 #pragma once
+#include <algorithm>
+#include <map>
 #include <string>
+#include <vector>
```
kontrybutor, line 5: This include is probably not needed as you do not use std::vector in your code.

WojtekMs: Yes, you are right.

ziobron:

PR #105, opened 2020-05-19
Fibonacci

WojtekMs:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,33 @@
...
+    if (sequence < 0) {
+        return -1;
+    }
+    if (sequence == 0) {
+        return 0;
+    }
+    if (sequence == 1) {
+        return 1;
+    }
+    if (sequence >= 2) {
```
jakjan95, line 28: Good error handling for negative number :) 
For 0 and 1 sequence you can make one condition, for recursive returning you don't need condition - this will happens if you don't get any previous conditions.

ziobron:

PR #116, opened 2020-05-20
Fibonacci

adrianmasiewicz:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,24 @@
...
+    }
+
+    return result;
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if (sequence == 0) return 0;
+    if (sequence == 1) return 1;
```
Marcinati, line 21: You can use simpler solution:
if (sequence < 2) return sequence;
All of your commits name are "adding works", It's misleading a bit. ;)

ziobron:

PR #117, opened 2020-05-20
Fibonacci ms00780

ms00780:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,18 @@
...
+      }
+  return tab[sequence];
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+  if (sequence == 1 or sequence == 0)
+    return sequence;
+  else
```
kuchichan, line 16: Else here is redundant, since when `if` will be fulfilled, we will leave the function with 1st `return`

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,18 @@
...
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+  int tab[sequence];
+    tab[0] = 0;
+    tab[1] = 1;
+      for (int i=2;i<=sequence;i++){
+        tab[i] = tab[i-1] + tab[i-2];
+      }
+  return tab[sequence];
```
mariomoore, line 10: In my opinion this code works fine, but i think you don't need to declare and fill the entire table with data, if you need just to return only last value of that table (but not whole table).

MechaniQ1: I agree with mario. When 'sequence' increases, the amount of memory needed in this function also increases. You can write this function with 3 variables only :)

ziobron:

PR #118, opened 2020-05-20
Calculate

MateuszKwiecien:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,18 @@
...
+    if(command == "subtract")
+        return std::to_string(first-second);
+    if(command == "multiply")
+        return std::to_string(first*second);
+    if(command == "divide")
+    {
+        if(second==0)
+            return "Division by 0";
+        else
+            return std::to_string(first/second);
```
mariomoore, line 16: This works fine, but you don't need `else` here, but:
```
if(second == 0)
    return "Division by 0";
return std::to_string(first / second);
```

DamiXioR: i agree with my predecessor, but in my opinion in this way it's better to use curly braces with if statement

nauka-programowania-MA: Yes, when someone is redundant we don't write this :)
Programmers are lazy :) try to write only what you really need.

ziobron:

PR #121, opened 2020-05-21
Powinno byc ok

Yoka40:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,27 @@
...
+      ret = std::to_string(first*second);
+    }
+  else if(command=="divide")
+    {
+      if(second == 0)
+      { ret = "Division by 0"; }
+      else
+      { ret = std::to_string(first/second);}
+    }
+  return ret;
```
MechaniQ1, line 27: I think instead of writing
`return ret;`
you can write returns in if/else if e.g.
`
  if(command=="add")
    {
     return std::to_string(first+second);
    }
`

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,27 @@
...
+  else if(command=="multiply")
+    {
+      ret = std::to_string(first*second);
+    }
+  else if(command=="divide")
+    {
+      if(second == 0)
+      { ret = "Division by 0"; }
+      else
+      { ret = std::to_string(first/second);}
```
DamiXioR, line 25: i think you can skip last else, it's unnecessary

ziobron:
All points granted in #127

PR #122, opened 2020-05-21
Calculate

Tomonike:
Working program Calculate


Tomonike:
> You named directory "bulid" which is not ignoring build files and add all of theme to repo. You should call it "build". I had same problem.

I know. I shold add this folder to ignored.

ziobron:

PR #123, opened 2020-05-21
adding works

smotan:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,21 @@
...
+        result = std::to_string(first + second);
+    } else if(command == "subtract") {
+        result = std::to_string(first - second);
+    } else if(command == "multiply") {
+        result = std::to_string(first * second);
+    } else if(command == "divide") {
+        if(second == 0) {
+            result = "Division by 0";
+        } else {
+            result = std::to_string(first / second);
```
DamiXioR, line 16: in my opinion else is redundant

nauka-programowania-MA: else will be redundant when in if statement will be return function. 
In this case when you remove else, result will always be  `std::to_string(first / second);`

ziobron:

PR #127, opened 2020-05-22
fibb&calc

Yoka40:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,27 @@
...
+  else if(command=="multiply")
+    {
+      ret = std::to_string(first*second);
+    }
+  else if(command=="divide")
+    {
+      if(second == 0)
+      { ret = "Division by 0"; }
+      else
+      { ret = std::to_string(first/second);}
```
DamiXioR, line 25: hi, think this else is unnecessary

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,25 @@
...
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+  if(sequence==0)  return 0;
+
+  else if(sequence<3)  return 1;
+  
+  else
```
DamiXioR, line 23: only sequence => 1 should return sequence, for 2 it's redundant, also you can skip last else

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,27 @@
 #include <string>
 
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+
+ std::string ret = "Invalid data";
+ //jak zrobić warunek na int ?
+  if(command=="add")
```
piezol, line 8: Formatting of curly braces does not adhere to any standard i know of (and is inconsistent even within file).

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,25 @@
 #pragma once
 
+
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    
+    int fibonacci[sequence];
```
piezol, line 6: If you use a c-style array with size that is unknown at the compile time, you HAVE to allocate memory dynamically. 

Also, for the iterative algorithm any kind of array is not needed - you only ever need the last two numbers, so at the best case you would 'only' use too much memory here.

module1/homework/calculate/calculate.hpp
ziobron: It's needed because there is no return in if.

ziobron:

PR #129, opened 2020-05-22
Homework2

Kuba-Wi:
Second pull request. Instead to branch master, this time to branch module2. Like it was said in the homework description.


module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,12 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sumEven = 0;
+    for(const auto& i : numbers) {
```
Kasperice, line 6: I think that you can use different name than "i"

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,13 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    int max;
```
Kasperice, line 6: You can assing value to ```max``` in the same line

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,29 @@
...
+        rhs = mod;
+    }
+    return lhs;
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    int nwd = NWD(lhs, rhs);
+    if(!nwd) {
```
piezol, line 24: Negation of integer implicitly cast to boolean uhhh :) 
Comparing to zero is much clearer.

module2/task1.cpp
```diff
@@ -2,8 +2,10 @@
 #include <vector>
 #include <string>
 
-// Implement printVec to print the content of a given vector on screen.
-// Each string in a new line.
+void printVector(const std::vector<std::string> & vec) {
```
piezol, line 5: As far as i remember we are not supposed to include task* files in pull requests - they are not a part of a homework.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,29 @@
...
-    // TODO: Implement me :)
-    return -1;
+    int mod;
+
+    if(lhs < 0) {
+        lhs = -lhs;
+    }
+    if(rhs < 0) {
+        rhs = -rhs;
+    }
```
anraf1001, line 12: You can use `std::abs` instead of that conditions
```suggestion
    lhs = std::abs(lhs);
    rhs = std::abs(rhs);
```

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,29 @@
...
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    int nwd = NWD(lhs, rhs);
+    if(!nwd) {
+        return 0;
+    }
+
+    return abs(lhs * rhs) / nwd;
```
anraf1001, line 28: You don't even need `nwd` variable. NWD is only equals 0 if min. one of the input variables equals 0.
```suggestion
    if(lhs == 0 || rhs == 0) {
        return 0;
    }

    return abs(lhs * rhs) / NWD(lhs, rhs);
```

piezol: I'm copying my comment from different PR:

module2/homework/addEven/addEven.hpp
nauka-programowania-MA: definitely, num is mach better name

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> sequence;
+    int add = step;
```
nauka-programowania-MA, line 6: Why? You have this value already in step

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
...
+
+    for(int i = 0; i < count; ++i) {
+        sequence.push_back(step);
+        step += add;
+    }
+
+    return sequence;
```
nauka-programowania-MA, line 13: maybe 

for (int i = 1 ; i <= count ; ++i) {
  sequence.push_back(step * i);
}

module2/homework/nwdNww/nwdNww.hpp
nauka-programowania-MA: agree

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,29 @@
...
+    return lhs;
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    int nwd = NWD(lhs, rhs);
+    if(!nwd) {
+        return 0;
+    }
```
nauka-programowania-MA, line 26: you could simplify to 
if (nwd == 0)
  return 0;

avoid implicit cast unsigned integer to bool

module2/task1.cpp
Kuba-Wi: How not to include task* files if they were already in branch module2?

piezol: You specify which files you add to commit by 'git add'.

Kuba-Wi: But before I made branch homework I commited task files on branch module2.

piezol: 'I commited task files on branch module2' - yeah, you should create a separate branch if you wanted to resolve those tasks within the repository. You can avoid such issues in future by doing so (it's ok as it is here, those PR aren't actually pulled, but you should try avoid that in the future).

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence

PR #130, opened 2020-05-22
Homework Podstawy C++ #2

kozlofski:
Author @kozlofski

### DIFF
--- a/module2/homework/addEven/addEven.hpp
+++ b/module2/homework/addEven/addEven.hpp
@@ -2,7 +2,14 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    if(numbers.size() == 0) {
+        return 0;
+    }
+    int sumOfEvenNumbers = 0;
+    for(auto elem : numbers) {
+        if(!(elem & 0x1)) {
+            sumOfEvenNumbers += elem;
+        }
+    }
+    return sumOfEvenNumbers;
 }
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -2,6 +2,16 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    if(count < 1 || step < 1) {
+        return {};
+    }
+    int number = step;
+    std::vector<int> sequence;
+
+    for(int i = 1; i <= count; i++) {
+        number = step * i;
+        sequence.push_back(number);
+    }
+
+    return sequence;
 }
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -2,7 +2,7 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    if(count < 1 || step < 1) {
+    if(count < 1) {
         return {};
     }
     int number = step;
--- a/module2/homework/maxOfVector/maxOfVector.hpp
+++ b/module2/homework/maxOfVector/maxOfVector.hpp
@@ -2,7 +2,16 @@
 #include <vector>
 #include <limits>
 
-int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+int maxOfVector(const std::vector<int>& vec) {    
+    if(vec.size() == 0) {
+        return 0;
+    }
+    int max = std::numeric_limits<int>::min();
+
+    for(auto elem : vec) {
+        if(elem > max) {
+            max = elem;
+        }
+    }
+    return max;
 }
--- a/module2/homework/nwdNww/main.cpp
+++ b/module2/homework/nwdNww/main.cpp
@@ -5,5 +5,11 @@ int main() {
   std::cout << "NWW(255, 15) = " << NWW(255, 15) << "\n";
   std::cout << "NWD(255, 15) = " << NWD(255, 15) << "\n";
 
+
+  std::cout << "NWW(255, 15) = " << NWW(
+    0.5 * (std::numeric_limits<int>::max() - 1),
+    std::numeric_limits<int>::max() - 1) << "\n";
+  
+
   return 0;
 }
--- a/module2/homework/nwdNww/nwdNww.hpp
+++ b/module2/homework/nwdNww/nwdNww.hpp
@@ -1,11 +1,36 @@
 #pragma once
+#include <limits>
 
 int NWD(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    if(lhs * rhs == 0) {
+        return lhs + rhs;
+        // if one of them is 0, return bigger of them
+    }
+    int nwd = std::abs(lhs);
+    lhs = std::abs(lhs);
+    rhs = std::abs(rhs);
+    while(lhs != rhs) {
+        nwd = abs(lhs - rhs);
+        lhs > rhs ? lhs = nwd : rhs = nwd;
+    }
+    return nwd; 
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    // Function should return int,
+    // so inputing to large values can result 
+    // in int overflow.
+    // Testing their product must be performed on int types
+    // two times bigger
+    if(lhs == 0 || rhs == 0) {
+        return 0;
+    }
+    int64_t product = std::abs
+        ( static_cast<int64_t>(lhs) * static_cast<int64_t>(rhs) ) 
+            / NWD(lhs, rhs) ;
+
+    if(product > std::numeric_limits<int>::max()) {
+        return 0;
+    }
+    return static_cast<int>(product);
 }
--- a/module2/homework/nwdNww/test.cpp
+++ b/module2/homework/nwdNww/test.cpp
@@ -1,4 +1,5 @@
 #include <vector>
+#include <limits>
 
 #include "gtest/gtest.h"
 #include "nwdNww.hpp"
@@ -67,8 +68,22 @@ INSTANTIATE_TEST_SUITE_P(
         std::make_tuple(1, -1, 1),
         std::make_tuple(-1, -1, 1),
         std::make_tuple(-10, -10, 10),
-        std::make_tuple(-10, -2, 10)
-        
+        std::make_tuple(-10, -2, 10),
+
+        std::make_tuple(std::numeric_limits<int>::max() - 1, 
+        std::numeric_limits<int>::max() - 2, 0),
+
+        // above test will perform 2^31 iterations in NWD() !!
+        // on my computer it took 9.92 s for performing all tests :)
+
+        std::make_tuple(0.5 * (std::numeric_limits<int>::max() - 1), 
+        std::numeric_limits<int>::max() - 1, 
+        std::numeric_limits<int>::max() - 1)
+        // the product of two large ints may result in int overflow,
+        // but function nwdNww() operates internally on long ints
+        // and if the result is smaller than MAX_INT
+        // it returns proper result
         // based on implementation of std::lcm
+        // Else, function returns 0.
     )
 );


module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,14 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    if(numbers.size() == 0) {
```
piezol, line 5: Edge case condition is not needed here - your loop would simply not go through any element for empty vector

module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,14 @@
...
+        return 0;
+    }
+    int sumOfEvenNumbers = 0;
+    for(auto elem : numbers) {
+        if(!(elem & 0x1)) {
```
piezol, line 10: :)

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -2,7 +2,16 @@
 #include <vector>
 #include <limits>
 
-int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+int maxOfVector(const std::vector<int>& vec) {    
+    if(vec.size() == 0) {
```
piezol, line 6: Behaviour is unspecified for empty vector both by the task description and by test cases. I would intentionally fail in case of empty vector being passed, since the statement '0 is the biggest number in empty vector' is obviously false.

module2/homework/nwdNww/main.cpp
```diff
@@ -5,5 +5,11 @@ int main() {
   std::cout << "NWW(255, 15) = " << NWW(255, 15) << "\n";
   std::cout << "NWD(255, 15) = " << NWD(255, 15) << "\n";
 
+
+  std::cout << "NWW(255, 15) = " << NWW(
```
piezol, line 9: We should not commit any code to main.cpp files - we only make changes to specified functions.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,36 @@
...
+    if(lhs * rhs == 0) {
+        return lhs + rhs;
+        // if one of them is 0, return bigger of them
+    }
+    int nwd = std::abs(lhs);
+    lhs = std::abs(lhs);
+    rhs = std::abs(rhs);
+    while(lhs != rhs) {
+        nwd = abs(lhs - rhs);
+        lhs > rhs ? lhs = nwd : rhs = nwd;
```
piezol, line 14: I would argue that free-hanging expressions that are not calls of void functions are ugly, but that is opinion-based i guess.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,36 @@
...
+    // Function should return int,
+    // so inputing to large values can result 
+    // in int overflow.
+    // Testing their product must be performed on int types
+    // two times bigger
+    if(lhs == 0 || rhs == 0) {
+        return 0;
+    }
+    int64_t product = std::abs
+        ( static_cast<int64_t>(lhs) * static_cast<int64_t>(rhs) )
```
piezol, line 29: Casting of just one would be fine.

module2/homework/nwdNww/test.cpp
```diff
@@ -67,8 +68,22 @@ INSTANTIATE_TEST_SUITE_P(
         std::make_tuple(1, -1, 1),
         std::make_tuple(-1, -1, 1),
         std::make_tuple(-10, -10, 10),
-        std::make_tuple(-10, -2, 10)
```
piezol, line 70: I think any additional tests we'd like to see should go to separate pull requests.

anraf1001: It should be the same PR like it is.

piezol: @anraf1001 Yes, I stand corrected.

module2/homework/addEven/addEven.hpp
Rosein, line 10: We should avoid magic numbers in code, so I suggest to use some const or macro here.

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,16 @@
...
-    // TODO: Implement me :)
-    return {};
+    if(count < 1) {
+        return {};
+    }
+    int number = step;
+    std::vector<int> sequence;
+
+    for(int i = 1; i <= count; i++) {
+        number = step * i;
```
Rosein, line 12: You can achieve the same result without multiplying, but with adding and it could be more efficient. Nevertheless these two lines are correct!

module2/homework/addEven/addEven.hpp
piezol: @Rosein
I do not agree. There is no way to informatively name a constant here.
Constant named 'one' is not better than using magic number.

Rule of thumb - if the only name you can think of for the constant is its exact value as read in english, just use the value.
const int one = 1; -- makes no sense
const int min_length_of_password = 10; -- makes sense

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,16 @@
...
+        sequence.push_back(number);
```
Rosein, line 13: It is correct, but cost is high. Read about behaviour of vector, when it reaches its end and run push_back(). I recommend prepare a vector with correct size before loop.

module2/homework/addEven/addEven.hpp
anraf1001: @Rosein
You should always specify which line you're talking about. If you have any suggestion about line 8, choose only line 8.

module2/homework/generateSequence/generateSequence.hpp
anraf1001: OT Alert
@Rosein
You can create as many comments as you want in one review. You don't have to create many independent reviews.

module2/homework/maxOfVector/maxOfVector.hpp
Rosein, line 6: There is a special method for checking wheter vector is empty. I agree with piezol about returning 0 as a misleading info for empty vector. Maybe some exception?

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -2,7 +2,16 @@
...
+        return 0;
+    }
+    int max = std::numeric_limits<int>::min();
+
+    for(auto elem : vec) {
+        if(elem > max) {
+            max = elem;
+        }
+    }
```
Rosein, line 15: Curly brackets here are unnecessary. You can write single instruction after if or for without any curly brackets.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,36 @@
...
+    // two times bigger
+    if(lhs == 0 || rhs == 0) {
+        return 0;
+    }
+    int64_t product = std::abs
+        ( static_cast<int64_t>(lhs) * static_cast<int64_t>(rhs) ) 
+            / NWD(lhs, rhs) ;
+
+    if(product > std::numeric_limits<int>::max()) {
+        return 0;
```
Rosein, line 33: Throw exception? E.g overflow?

module2/homework/addEven/addEven.hpp
Rosein: I have meant 10th line.  Usually when there is flag checking I meet something like 
#define IS_EVEN_FLAG 1
Maybe I am wrong.

module2/homework/generateSequence/generateSequence.hpp
Rosein: What do you mean by OT?

anraf1001: Off-Topic ;-)

module2/homework/addEven/addEven.hpp
anraf1001: I think that @kozlofski could just use modulo operator.

module2/homework/maxOfVector/maxOfVector.hpp
anraf1001: No. He shouldn't do that. You should always use curly brackets even if there is only 1 line.

module2/homework/addEven/addEven.hpp
Rosein: You are right it is also good another way.

kozlofski: I wrote it that way because I've heard that such a bitwise operation (checking one bit) would be much faster than division. Of course division by 2 is simple for a computer an probably (I don't know) is optimised by a compiler to simply shifting bits to the right - probably some test should be performed or we could somehow it watch the machine code ;)  Also I don't think it's a magical formula, it's quite basic bit operation.

module2/homework/maxOfVector/maxOfVector.hpp
kozlofski: I know I can write it without {} but I think adding brackets makes code easier to maintain - for example if I would like to add some other instructions here.

Rosein: You are right, using here curly braces is good practice. My apologise.

module2/homework/addEven/addEven.hpp
kozlofski: Yes, but in that case a variable is initiated unnecessarily, and I think the same happens for an unused loop.

module2/homework/addEven/addEven.hpp
piezol: @kozlofski
1) Bitwise and is operation simple enough so you don't have to justify yourself for using it here IMO
2) Yes, 'bitwise and' will be much faster than division, but if you use "variable % 2" and 2 is compile time constant, compiler will avoid division at all. If variable is unsigned int, "%2" is the same operation as "&1" and is compiled to the same machine code with -O3 option.
With variable not being unsigned it's equivalent a little more complicated, but division is avoided.
3) Don't overthink those tasks performance-wise - it's not the point :)

module2/homework/maxOfVector/maxOfVector.hpp
kozlofski: isEmpty() method would be better here?

Rosein: empty() would be fine. But... Is zero really max value of an empty vector?

module2/homework/nwdNww/nwdNww.hpp
kozlofski: Yes, but I have no experience in exceptions in C++ :) I have only a little exp in that field in Java.

module2/homework/addEven/addEven.hpp
nauka-programowania-MA: Agree with @kozlofski exit form function if condition is not satisfy is better

module2/homework/addEven/addEven.hpp
nauka-programowania-MA, line 10: Modulo operator make this for you, you can simplyfy it :)
instead check bit, you can simply use modulo here.

if (num % 2

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA: Multiply is one machine function like adding. And using multiply, you don't need to store step value, so finally you use less memory and have the same efficient.

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA, line 13: You know how many value will be in vector. So you avoid few unnecessary memory allocation :)
use vec.reseerve(count)

module2/homework/maxOfVector/maxOfVector.hpp
nauka-programowania-MA: if vec.size() == 0 you can return what you want. It was not specify, what should be return for empty vec. 
Better option will be return std::optional<int> and if vec will be empty you will return std::nullopt.

But we will be learn std::optional on modern C++ course

nauka-programowania-MA: Try avoid exception

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -2,7 +2,16 @@
...
 #include <limits>
 
-int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+int maxOfVector(const std::vector<int>& vec) {    
+    if(vec.size() == 0) {
+        return 0;
+    }
+    int max = std::numeric_limits<int>::min();
```
nauka-programowania-MA, line 9: you can just use 
int max = vec[0]

module2/homework/nwdNww/test.cpp
```diff
@@ -67,8 +68,22 @@ INSTANTIATE_TEST_SUITE_P(
...
+        std::numeric_limits<int>::max() - 1, 
+        std::numeric_limits<int>::max() - 1)
+        // the product of two large ints may result in int overflow,
+        // but function nwdNww() operates internally on long ints
+        // and if the result is smaller than MAX_INT
+        // it returns proper result
         // based on implementation of std::lcm
+        // Else, function returns 0.
     )
 );
```
nauka-programowania-MA, line 89: add new line at the end

module2/homework/addEven/addEven.hpp
ziobron: You can write your own performance tests: http://quick-bench.com/IJej-E3IJB5Ewsw2oh22W1rYxZA
Modulo operation is usually more clear for people.

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence

PR #131, opened 2020-05-22
Homework Podstawy C++ #2

anraf1001:
Authors: @Vicx95, @anraf1001

### DIFF
--- a/module2/homework/addEven/addEven.hpp
+++ b/module2/homework/addEven/addEven.hpp
@@ -2,7 +2,13 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+
+    for (const auto& num : numbers) {
+        if (num % 2 == 0) {
+            sum += num;
+        }
+    }
+
+    return sum;
 }
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -3,21 +3,16 @@
 
 std::vector<int> generateSequence(int count, int step) {
     std::vector<int> sequence;
-    if(count <= 0)
-    {
+    if (count <= 0) {
         return sequence;
     }
-    for(int i = 0; i < count; i++)
-    {
-        if(i == 0)
-        {
+    for (int i = 0; i < count; i++) {
+        if (i == 0) {
             sequence.push_back(step);
-        }
-        else
-        {
+        } else {
             sequence.push_back(sequence[i - 1] + step);
         }
     }
-    
+
     return sequence;
 }
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -3,15 +3,14 @@
 
 std::vector<int> generateSequence(int count, int step) {
     std::vector<int> sequence;
+
     if (count <= 0) {
         return sequence;
     }
-    for (int i = 0; i < count; i++) {
-        if (i == 0) {
-            sequence.push_back(step);
-        } else {
-            sequence.push_back(sequence[i - 1] + step);
-        }
+
+    sequence.push_back(step);
+    for (int i = 1; i < count; i++) {
+        sequence.push_back(sequence[i - 1] + step);
     }
 
     return sequence;
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -8,9 +8,11 @@ std::vector<int> generateSequence(int count, int step) {
         return sequence;
     }
 
-    sequence.push_back(step);
+    sequence.resize(count);
+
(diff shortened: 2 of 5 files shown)


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,22 @@
...
+    for(int i = 0; i < count; i++)
+    {
+        if(i == 0)
+        {
+            sequence.push_back(step);
+        }
+        else
+        {
+            sequence.push_back(sequence[i - 1] + step);
+        }
```
Kasperice, line 19: If you rebuild that loop and start from 1 to `i <= count` it will be possible to add correct elements to vector in one line only. But this is a comment how to do that in shorter way - your solution is also good.

Rosein: Curly brackets here are unnecessary. You can write single instruction after loop,if,else  without any curly brackets.
It is correct, but cost of push_back is high. Read about behaviour of vector, when it reaches its end and runs push_back(). I recommend prepare a vector with correct size before loop.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,19 @@
...
+    int maxValue = vec[0];
+    
+    for(const auto& elem : vec)
+    {
+        if(maxValue < elem)
+        {
+            maxValue = elem;
+        }
+    }
+    return maxValue;
```
Rosein, line 20: Curly brackets here are unnecessary.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,19 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    if(vec.empty())
+    {
+        return 0;
```
Rosein, line 8: Is zero really max value of an empty vector?

module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,13 @@
...
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+
+    for (const auto& num : numbers) {
+        if (num % 2 == 0) {
+            sum += num;
+        }
+    }
```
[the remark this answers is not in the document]
piezol: Additional curly brackets for single statements are not a bad practice.

module2/homework/maxOfVector/maxOfVector.hpp
anraf1001: You should always use curly brackets even if there is only 1 line

module2/homework/addEven/addEven.hpp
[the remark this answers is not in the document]
anraf1001: You should always use curly brackets even if there is only 1 line

module2/homework/maxOfVector/maxOfVector.hpp
piezol: @Rosein That's a good remark, but the behaviour is unspecified in task description, so returning 0 is not really a bug

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,23 @@
...
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    if (lhs == 0 || rhs == 0) {
+        return 0;
+    }
+
+    return std::abs(lhs * rhs) / NWD(lhs, rhs);
```
piezol, line 22: I'm copying my comment from different CR:


module2/homework/generateSequence/generateSequence.hpp
Rosein: I see @anraf1001, that you disagree with my opinion about curly braces. I have made a research in [Google Coding](https://google.github.io/styleguide/cppguide.html#Conditionals) standard and in fact you are right. There is no such recommendations about not using {} in single instructions blocks.

module2/homework/maxOfVector/maxOfVector.hpp
Rosein: Not always. But you are right, that using them is ok.

module2/homework/maxOfVector/maxOfVector.hpp
nauka-programowania-MA: if vec.size() == 0 you can return what you want. It was not specify, what should be return for empty vec.
Better option will be return std::optional and if vec will be empty you will return std::nullopt.

But we will be learn std::optional on modern C++ course

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence
+8 XP delivery before 31.05

{
  "users": ["Vicx95", "anraf1001"]
}

ziobron:
{
"requirement_ids": [1636],
"points": [3],
"users": ["Vicx95", "anraf1001"]
}

PR #132, opened 2020-05-22
adding works

Maciej3206:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,16 @@
...
+
+std::string calculate(const std::string &command, int first, int second) {
+    if (command == "add") return std::to_string(first + second);
+    else if (command == "subtract") return std::to_string(first - second);
+    else if (command == "multiply") return std::to_string(first * second);
+    else if (command == "divide"){
+        if (second==0) {
+            return ("Division by 0");
+        }
+        else return std::to_string(first / second);
```
anraf1001, line 13: This `else` is not needed.

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,16 @@
...
+    }
+    else return ("Invalid data");
```
anraf1001, line 15: This `else` is not needed.

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,16 @@
...
-    return "";
-}
+#pragma once
+#include <string>
+#include <iostream>
+
+std::string calculate(const std::string &command, int first, int second) {
+    if (command == "add") return std::to_string(first + second);
+    else if (command == "subtract") return std::to_string(first - second);
+    else if (command == "multiply") return std::to_string(first * second);
```
anraf1001, line 8: You should use `{}` in if/else statements even when there is only one line.

Rosein: I do not agree. Curly brackets in my opinion are unnecessary if you want write only single instruction. But... if and only if you write that instruction in new line with an extra intend.

anraf1001: Curly brackets make it easier to realize what line is connected with what if statement. It's good practice to always use curly brackets.

module1/homework/calculate/calculate.hpp
Maciej3206: Thank You, You heave right. I will make this change.

ziobron:

PR #133, opened 2020-05-22
adding works

kali-03:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,27 @@
 #include <string>
 
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+    if (command=="add"){
+        std::string add=std::to_string(first+second);
```
Kasperice, line 6: You can return result right away, without creating new variable (in all cases)

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,27 @@
...
+        return (multiply);
+    }
+    else if(command=="divide"){
+        if (second==0){
+        return ("Division by 0");
+        }else{
+        std::string divide=std::to_string(first/second);
+        return (divide);
+        }
+    }
```
Kasperice, line 24: It is not clear, please add some indentations. Try to use one convention with if's, sometime you add whitespace before bracket, sometimes not.

ziobron:

PR #134, opened 2020-05-22
Fibonacci

kali-03:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,38 @@
...
+
+    if (sequence==0)
     return 0;
+    
+    if (sequence==1) 
+    return 1;
+
+    else 
+    {
+        for (int i=2;i<=sequence;i++){
```
Kasperice, line 14: You are using two different conventions of opening brackets. Try using only one.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,38 @@
...
-    // TODO: Your implementation goes here
+    
+    if (sequence==0)
     return 0;
+
+    if (sequence==1)
+    return 1;
+
+    if (sequence==2)
+    return 1;
```
Kasperice, line 33: Please use brackets and indentation, it will be clear then.
`sequence == 0` and `sequence == 1` can be covered in one `if` when you return some variable.
`sequence == 2` is not necessary here, it will be covered by recursion of `fibonacci_recursive(0)+ fibonacci_recursive(1)`

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,38 @@
...
+
+    else
+    return fibonacci_recursive(sequence-1)+ fibonacci_recursive(sequence-2);
```
Kasperice, line 36: Missing indentation and maybe some brackets?

kali-03: Maybe but if u have only one line after conditional statements, you don't need to use brackets.

piezol: > 
> 
> Maybe but if you have only one line after conditional statement, you don't need to use brackets.

Yes, but at least use proper indentation.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,38 @@
...
     return 0;
+    
+    if (sequence==1) 
+    return 1;
+
+    else 
+    {
+        for (int i=2;i<=sequence;i++){
+        tabela[0]=0;
+        tabela[1]=1;
```
Rosein, line 16: There is no need to make these two instructions on every step of loop. Maybe take them outside?

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +0,0 @@
-#pragma once
-#include <string>
-
-std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
-}
```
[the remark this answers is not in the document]
kali-03: Next time i will do it. :)

module1/homework/fibonacci/fibonacci.hpp
kali-03: Good idea, next time i will improve it.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,38 @@
 #pragma once
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
+    int tabela[sequence];
```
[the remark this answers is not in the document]
piezol: This is unsafe code - static arrays should have size that is known at compile time, for a size that comes from a variable array must be dynamically allocated. 
But as Kasperice said - array is not necessary here in the first place.

ziobron:

PR #135, opened 2020-05-22
Calculate task

lisie-sprawy:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,21 @@
-#pragma once
 #include <string>
 
 std::string calculate(const std::string& command, int first, int second) {
     // TODO: Implement your solution here and return proper value
-    return "";
-}
+        if(command == "add"){
+        return std::to_string(first + second);
```
Kasperice, line 6: Some indentation will improve readability

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,21 @@
...
+        }
+        if(command == "multiply"){
+        return std::to_string(first * second);
+        }
+        if(command == "divide"){
+    	    if(second == 0){
+            return "Division by 0";
+            }
+        return std::to_string(first/second);
+         }
```
Kasperice, line 19: Inside that part indentation will be very important (of course it is working now), but I was wondering why `return std::to_string(first/second)` is called after `if` then I noticed it is inside.

`return std::to_string(first/second)` I saw that you are using whitespaces around operators, please add whitespaces also here

module1/homework/calculate/calculate.hpp
lisie-sprawy: Yeah, thank you very much for the comments, I'll work on readability of my code!

ziobron:

PR #137, opened 2020-05-23
Homework Podstawy C++ #2

Kasperice:
Authors: @Kasperice @Marcinati

### DIFF
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -2,7 +2,7 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    std::vector<int> result_vector = {};
+    std::vector<int> result_vector;
     if (count <= 0) {
         return result_vector;
     }
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -2,6 +2,12 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> result_vector = {};
+    if (count <= 0) {
+        return result_vector;
+    }
+    for (size_t i = 1; i <= count; i++) {
+        result_vector.push_back(i * step);
+    }
+    return result_vector;
 }
--- a/module2/homework/maxOfVector/maxOfVector.hpp
+++ b/module2/homework/maxOfVector/maxOfVector.hpp
@@ -3,8 +3,11 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
+    if (vec.empty()) {
+        return 0;
+    }
     int maxVec = vec[0];
-    for (size_t i = 0; i < vec.size(); ++i) {
+    for (size_t i = 1; i < vec.size(); ++i) {
         if (maxVec < vec[i]) {
             maxVec = vec[i];
         }
--- a/module2/homework/maxOfVector/maxOfVector.hpp
+++ b/module2/homework/maxOfVector/maxOfVector.hpp
@@ -3,6 +3,11 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    int maxVec = vec[0];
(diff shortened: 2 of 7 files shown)


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,12 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> result_vector = {};
```
anraf1001, line 5: `std::vector` is empty by default.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,11 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    int maxVec = vec[0];
+    for (size_t i = 0; i < vec.size(); ++i) {
```
anraf1001, line 7: You could start from index 1 because you've already used `vec[0]` last line.
```suggestion
    for (size_t i = 1; i < vec.size(); ++i) {
```

module2/homework/generateSequence/generateSequence.hpp
Kasperice: Thanks, I changed this

module2/homework/maxOfVector/maxOfVector.hpp
Kasperice: You are right, I did not notice it. It is changed now

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,12 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> result_vector;
+    if (count <= 0) {
```
piezol, line 6: You don't need that if statement - for count <= 0 your loop will simply not add any element, so you will return empty vector correctly anyway.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,15 @@
...
+    return std::abs(NWD(rhs % lhs, lhs)); 
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    if (lhs == 0 || rhs == 0) {
+        return 0;
+    }
+    return std::abs((lhs * rhs) / NWD(lhs, rhs));
```
piezol, line 14: I'm copying my comment from different PR:

Marcinati: I'm not sure if we need to allocate more memory here. We have tests made by our client for small numbers (and it's passed), in other way we will use everywhere long long variables (even in previous homework - fibonacci), but it costs us a memory. We don't know exactly what is more important here - memory or greater range of variables? @ziobron can you tell us more?

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,12 @@
...
+        return result_vector;
+    }
+    for (size_t i = 1; i <= count; i++) {
+        result_vector.push_back(i * step);
```
Rosein, line 10: It is correct, but cost is high. Read about behaviour of vector, when it reaches its end and run push_back(). I recommend prepare a vector with correct size before loop.

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA: use return result_vector

module2/homework/nwdNww/nwdNww.hpp
ziobron: it is ok now. You don't need to do anything extra. In future will pay more attention to details like this one.

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence
+8 XP delivery before 31.05

{
"users": ["Kasperice", "Marcinati "]
}

ziobron:
Automation didn't work for Marcinati, trying one more time

{
"users": ["Marcinati"]
}

ziobron:
Ok now :)

PR #138, opened 2020-05-23
all works

TomaszEichler:


TomaszEichler:
Yes I know, I didn't want to do this that way. I used git commit command with wrong flag

PR #139, opened 2020-05-23
Fibonacci

TomaszEichler:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,34 @@
...
-int fibonacci_iterative(int sequence) {
+int fibonacci_iterative(int sequence){
     // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0){
+        return 0;
+    }
+    else if(sequence == 1){
+        return 1;
+    }
```
mariomoore, line 10: It works fine, but you can also write it this way:
```
if (sequence < 2) {
    return sequence;
}
```

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,34 @@
...
-    // TODO: Your implementation goes here
-    return 0;
+int fibonacci_recursive(int sequence){
+
+    if(sequence == 0){
+        return 0;
+    }
+    else if(sequence == 1){
+        return 1;
+    }
```
mariomoore, line 31: What do you think about:
```
if (sequence < 2) {
    return sequence;
}
``` ?

module1/homework/fibonacci/fibonacci.hpp
TomaszEichler: Yes I know, I did it consciously. The most common description of this algorithm separates these two cases so I did the same. Anyway thank you :)

module1/homework/fibonacci/fibonacci.hpp
TomaszEichler: Same here :)

ziobron:

PR #141, opened 2020-05-23
Merged calculate and fibonacci

TomaszEichler:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -21,7 +21,14 @@ int fibonacci_iterative(int sequence) {
...
-    // TODO: Your implementation goes here
-    return 0;
+int fibonacci_recursive(int sequence){
+
+    if(sequence == 0){
+        return 0;
+    }
+    else if(sequence == 1){
+        return 1;
+    }
```
Pawellos, line 31: You could write in simple way, for example: 
if(sequence < 2)
   return sequence;

TomaszEichler: I know I could but I didn' do that consciously. The most common description of this algorithm separates these two cases so I did the same. Overall it doesn't matter, shortening everything have never been my goal, I rather like make it clean and readable

Pawellos: Sure, it was only suggestion, you are a developer and create that code, so it is your decision.

TomaszEichler: OK, thank you anyway! :) Feel free to review my future PRs.

PR #142, opened 2020-05-23
Homework2

mariomoore:


module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,29 @@
...
+    return higher;
 }
 
-int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+int NWW(int lhs, int rhs)
+{
+    int nwd = NWD(lhs, rhs);
+    if (nwd)
```
anraf1001, line 23: Implicit casting to boolean. Comparing to 0 is much clearer
```suggestion
    if (nwd != 0)
```

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,29 @@
...
+    {
+        return std::abs(lhs * rhs) / nwd;
```
Rosein, line 25: I'm copying piezol's comment from different PR:

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,20 @@
...
+{
+    if (count > 0)
+    {
+        std::vector<int> sequence(count);
+        int val2add = step;
+        for (auto& i : sequence)
+        {
+            i = val2add;
+            val2add += step;
+        }
```
nauka-programowania-MA, line 14: you can avoid val2add 
just use sequence * -> when you use for loop (int i.....)
or sequence[i] = sequence[i-1] + step -> but remember to assign value to sequence[0] and iterator form 1 to count

ziobron:

PR #143, opened 2020-05-23
Homework Podstawy C++ #2

jakjan95:
Authors: @jakjan95

### DIFF
--- a/module2/homework/addEven/addEven.hpp
+++ b/module2/homework/addEven/addEven.hpp
@@ -2,7 +2,12 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int result = 0;
+
+    for(const auto& el : numbers){
+        if( el % 2 == 0){
+            result += el;
+        }
+    }
+    return result;
 }
--- a/module2/homework/maxOfVector/maxOfVector.hpp
+++ b/module2/homework/maxOfVector/maxOfVector.hpp
@@ -3,6 +3,12 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    auto maximal = vec[0];
+    for(const auto& el : vec){
+        if(el > maximal){
+            maximal = el;
+        }
+    }
+
+    return maximal;
 }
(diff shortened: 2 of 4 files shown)


module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,12 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    auto maximal = vec[0];
```
Kasperice, line 6: We are passing vector of ints, so our maximal value will also be int

module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,12 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int result = 0;
+
+    for(const auto& el : numbers){
```
Kasperice, line 7: You can use full variable name instead of `el` it would be easier to understand in bigger code

jakjan95: Using variable name is common practice in python, Personally I prefer using el which is shortcut of element and it is universal for iteration over std containers, also during lesson Mateusz recommended using el for iterating over containers.

module2/homework/maxOfVector/maxOfVector.hpp
piezol: > 
> 
> We are passing vector of ints, so our maximal value will also be int

So using auto is not an error or bad practice here..

module2/homework/addEven/addEven.hpp
piezol: > 
> 
> You can use full variable name instead of `el` it would be easier to understand in bigger code

I agree with jakjan, for simple collection loops 'el' is as abvious as 'i' in regular for loops.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,16 @@
...
+    return NWD(rhs, lhs % rhs);
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    if(lhs == 0 || rhs == 0){
+        return 0;
+    }
+    return std::abs(lhs * rhs) / NWD(lhs, rhs);
```
Rosein, line 15: I'm copying piezol's comment from different PR:

ziobron:

PR #144, opened 2020-05-23
add fibonacci functions

alekczapski:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,37 @@
...
+
+    int results[sequence + 1];
+    results[0] = 0;
+    results[1] = 1;
+    results[2] = 1;
+
+    for (int i = 3; i <= sequence; ++i)
+    {
+        results[i] = results[i - 1] + results[i - 2];
+    }
```
TomaszEichler, line 22: It's OK, but you can also go with for loop, starting with `i = 2` when you change `else if` to ` else if (sequence == 1){
return 1;
}` for example , and remove `result[2] = 1;`.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,37 @@
...
+{
+    if (sequence == 0)
+    {
+        return 0;
+    }
+    else if (sequence <= 2)
+    {
+        return 1;
+    }
+    return fibonacci_recursive(sequence - 1) + fibonacci_recursive(sequence - 2);
```
Marcinati, line 36: You can simplify it:

if (sequence < 2) {
return sequence;
}

There is no need here to compare "2" in if condition, because your last line will return correct result for sequence = 2.

module1/homework/fibonacci/fibonacci.hpp
Pawellos: I am not sure is correct way to solve this problem. You are using array, so you have to allocate a lot of memory for that resolve.

piezol: int results[sequence + 1]; 
Unsafe code - static arrays must have size that is known at the compile time.
If it's variable, dynamic allocation should be used.
And as Pawellos said - array is not required for iterative approach, since we only ever need at most 2 last numbers in sequence.

ziobron:

PR #146, opened 2020-05-23
Fibonacci task

lisie-sprawy:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,21 @@
...
 
+
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
-}
+    if(sequence == 0)
+        return 0;
+    if(sequence == 1)
+        return 1;
```
Pawellos, line 19: You could write:
if(sequence < 2) 
    return sequence;

mariomoore, line 19: This code works fine, but you can simplify it by:
```
if (sequence < 2) {
    return sequence;
}
```

lisie-sprawy: Thank you very much! Your code can replace my code in line 16-19, am I understand correctly?

mariomoore: Yes. It's also good practise to use currly brackets {} always, even if there is only one-line instruction like `return something;`

lisie-sprawy: good to know! thx:)

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,21 @@
 #pragma once
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    int a = 0, b = 1, c = 1;
```
Marcinati, line 4: Try to name variables depending of its purpose, (a, b, c, x, y, z... is't a good practice, because we don't know exactly what it is).

ziobron:

PR #149, opened 2020-05-23
Homework Podstawy C++ #2

MarekJanek:
Authors: @LogarithmDerivative, @MarekJanek

### DIFF
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -1,7 +1,19 @@
 #pragma once
 #include <vector>
 
-std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
+std::vector<int> generateSequence(int count, int step)
+{
+    if(count > 0)
+    {
+        std::vector<int> result;
+        int value = step;
+
+        for(size_t i = 0; i < count; i++)
+        {
+            result.push_back(value);
+            value += step;
+        }
+        return result;
+    }
     return {};
 }
--- a/module2/homework/maxOfVector/maxOfVector.hpp
+++ b/module2/homework/maxOfVector/maxOfVector.hpp
@@ -6,7 +6,7 @@ int maxOfVector(const std::vector<int>& vec) {
     
     int max=std::numeric_limits<int>::min();
     
-    for(int element : vec)
+    for(const int& element : vec)
     {
         if(element>max)
             max=element;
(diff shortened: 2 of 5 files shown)


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,19 @@
...
+std::vector<int> generateSequence(int count, int step)
+{
+    if(count > 0)
+    {
+        std::vector<int> result;
+        int value = step;
+
+        for(size_t i = 0; i < count; i++)
+        {
+            result.push_back(value);
```
piezol, line 13: Memory will have to be allocated several times.
The size of result vector is known beforehend (== count), so we should allocate memory once using reserve

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,13 @@
...
-    return {};
+    
+    int max=std::numeric_limits<int>::min();
+    
+    for(const int& element : vec)
+    {
+        if(element>max)
+            max=element;
+    }
+    return max;
```
piezol, line 14: That has defect of returning std::numberic_limits<int>::min() for empty vector.
But the behaviour was unspecified for that case, so i guess anything works.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,36 @@
...
-    // TODO: Implement me :)
-    return -1;
+
+    if(lhs==0 && rhs==0)
+        return 0;
+    
+    lhs=abs(lhs);
+    rhs=abs(rhs); 
+
+    return (lhs*rhs)/NWD(lhs,rhs);
```
piezol, line 34: There are trivial cases that your code won't work for, ie NWW(1e8, 1e8).

module2/homework/generateSequence/generateSequence.hpp
Rosein: Or instead using reserve you can set well size of vector at its creation.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,13 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    
+    int max=std::numeric_limits<int>::min();
```
nauka-programowania-MA, line 7: beter use max = vec[0]

ziobron:

PR #153, opened 2020-05-24
Homework Podstawy C++ #2

kszytko:
Authors: @Nessy90, @kszytko


module2/homework/addEven/addEven.hpp
```diff
@@ -1,8 +1,14 @@
...
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+int addEven(const std::vector<int>& numbers){
+
+    int result = 0;
+
+    for(const auto& element : numbers){
+   	 if (element % 2 == 0){
+	 result += element;
```
piezol, line 10: Line 10 requires indendation.

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,14 @@
 #pragma once
 #include <vector>
 
-std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+std::vector<int> generateSequence(const int count, const int step) {
+    std::vector<int> result;
```
piezol, line 5: Vec might get its memory allocated several times.
We know the size of vector upfront, so we should allocate memory for all elements with reserve function or with constructor.

module2/homework/generateSequence/main.cpp
```diff
@@ -2,7 +2,7 @@
 #include "generateSequence.hpp"
 
 int main() {
-    for (const int element : generateSequence(10, 5))
+    for (const int element : generateSequence(-3, 0))
```
piezol, line 5: Do not commit changes to main.cpp files.
Changing them is not part of our task.

module2/homework/nwdNww/main.cpp
```diff
@@ -2,7 +2,7 @@
 #include "nwdNww.hpp"
 
 int main() {
-  std::cout << "NWW(255, 15) = " << NWW(255, 15) << "\n";
+  std::cout << "NWW(255, 15) = " << NWW(0, 15) << "\n";
```
piezol, line 5: See above comment about main files

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,27 @@
...
+    
+    return first;
 }
 
-int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+int NWW(const int lhs, const int rhs) {
+    if(lhs && rhs){
+        return abs(lhs * rhs) / NWD(lhs, rhs);
```
piezol, line 24: I'm copying my comment from different PR:

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,27 @@
 #pragma once
+#include <string>
 
-int NWD(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+int NWD(const int lhs, const int rhs) {
```
piezol, line 4: You've changed the header of the function (i guess we were not supposed to do that).

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,14 @@
 #pragma once
 #include <vector>
 
-std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+std::vector<int> generateSequence(const int count, const int step) {
```
nauka-programowania-MA, line 4: const not needed

kszytko: So when is good to use const?

nauka-programowania-MA: When you pass value by reference. Or when you pass value as pointer, but you want make sure, noone modify it. When you pass copy like in this case, you don't care about it value, because even if you modify it only copy.

ziobron:
{
"users": ["Nessy90", "kszytko"]
}

ziobron:
+ points for cooperation

{
"requirement_ids": [1636],
"points": [3],
"users": ["Nessy90", "kszytko"]
}

PR #154, opened 2020-05-24
Mod2 complete homework

mariuszlisowski:


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,11 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+	std::vector<int> vec;
+	if (count > 0) {
+		int value {};
+		while (count--)
+			vec.push_back(value += step);
```
piezol, line 9: Vec might get its memory allocated several times.
We know the size of vector upfront, so we should allocate memory for all elements with reserve function or with constructor.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,18 @@
...
+	}
+	return abs(lhs);
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    if (!lhs || !rhs)
+		return 0;
+	return abs(lhs * rhs / NWD(lhs, rhs));
```
piezol, line 17: I'm copying my comment from different PR:

piezol: Namespace is not specified for abs (you should use std::abs)

module2/homework/generateSequence/generateSequence.hpp
mariuszlisowski: Yes, indeed. It couldn't be done with the constructor as 'count' can be negative so I used 'reserve' function as suggested (after making sure 'count' was positive). I also tried to use the access operator [] but it doesn't increase the size of a vector (although worked well with the 'resize' function). Thank you piezol!

module2/homework/nwdNww/nwdNww.hpp
mariuszlisowski: I overlooked that. Thanks for review piezol! Casting used so it shouldn't overflow now. Also appropriate 'absolute' function used for 'long long' (std::llabs). Tested for (1e8, 1e8) but still fails for (1e8-1, 1e8+1) as the function returns int.

```cpp
#pragma once
#include <cstdlib>

int NWD(int lhs, int rhs) {
    int remainder;
    while (rhs !=0) {
        remainder = lhs % rhs;
        lhs = rhs;
        rhs = remainder;
    }
    return std::abs(lhs);
}

int NWW(int lhs, int rhs) {
    if (!lhs || !rhs)
        return 0;
    return std::llabs(static_cast<unsigned long long>(lhs)
                      * rhs / NWD(lhs, rhs));
}
```

module2/homework/generateSequence/generateSequence.hpp
mariuszlisowski:
```cpp
#pragma once
#include <vector>

    std::vector<int> vec;
    if (count > 0) {
        vec.reserve(count);
        int value {};
        while (count--)
            vec.push_back(value += step);
    }
    return vec;
}
```

PR #155, opened 2020-05-24
Homework2

MonikaZelechowska:


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,9 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> generatedVec {};
```
piezol, line 5: We know the desired size of vector upfront so we should allocate memory in constructor or by using reserve() function.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,18 @@
...
+            max_divider = i;
+    }
+    return max_divider;
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    if (NWD(lhs, rhs)!= 0)
+        return abs(lhs*rhs)/NWD(lhs, rhs);
```
piezol, line 15: I'm copying my comment from different PR:

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,18 @@
...
+    else 
+        return abs(lhs*rhs);
```
piezol, line 17: Which would always be equal to 0, so we can return 0 :)

module2/homework/nwdNww/nwdNww.hpp
piezol: Also, std::abs instead of abs

ziobron:
{
"points": [2, 4, 2, 6, 2, 5, 2, 5]
}

PR #156, opened 2020-05-24
Homework Podstawy C++ #2

Petrus9107:


module2/homework/addEven/addEven.hpp
```diff
@@ -1,8 +1,22 @@
 #pragma once
+#include <iostream>
 #include <vector>
+#include <numeric>
+
+
+bool checkIfEven(int value){
+    return value%2 ==0 ;
```
piezol, line 8:
```suggestion
    return value % 2 == 0 ;
```
Spaces

module2/homework/addEven/addEven.hpp
```diff
@@ -1,8 +1,22 @@
...
+}
+
+int myBinaryPredicate(int a, int b){
+    return checkIfEven(b) ? a+b : a;
```
piezol, line 12:
```suggestion
    return checkIfEven(b) ? a + b : a;
```

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,18 @@
...
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+   
+    if (count <=0 )
+        return {};
+
+    std::vector<int> tmpVec;
+    tmpVec.reserve(count);
+    int initialValue= 0;
```
piezol, line 11: InitialValue is not a good name for this variable - it is changed in every iteration of loop, so it's not really 'initial' after first iteration.

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,18 @@
...
+
+    for (size_t i=0; i<count; i++)
```
piezol, line 13:
```suggestion
    for (size_t i=0; i < count; i++)
```

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -1,8 +1,10 @@
 #pragma once
 #include <vector>
 #include <limits>
+#include <set>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    std::set<int> mySet(vec.begin(), vec.end());
+    std::set<int>::iterator it= std::prev(mySet.end());
```
piezol, line 8:
```suggestion
    std::set<int>::iterator it = std::prev(mySet.end());
```

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -1,8 +1,10 @@
...
+    return (*it);
```
piezol, line 9: I believe the purpose of the task was to implement 'max' without relying on STL too much.
Also, 'naive' max implementation would have linear time complexity, while your has n * log(n) complexity, because you create set, which is pretty much as expensive as sorting array would be.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,17 @@
...
 
-int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+int NWW(int lhs, int rhs)
+{
+    if (lhs == 0 || rhs == 0) {
+        return 0;
+    }
+    return std::abs(lhs * rhs) / NWD(lhs, rhs);
```
piezol, line 16: I'm copying my comment from different PR:

ziobron:

PR #157, opened 2020-05-24
Homework 2

McKucia:


module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,9 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+        int sum=0;
+    for(const auto& el : numbers)
+        if( el%2 == 0)
```
piezol, line 7:
```suggestion
        if(el%2 == 0)
```

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,12 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    if(count <=0)
+    	return {};
+    std::vector<int> vec; 
+    vec.reserve(count); 
+    for(size_t i=1; i<=count; i++)
```
piezol, line 9:
```suggestion
    for(size_t i=1; i <= count; i++)
```

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,15 @@
...
+        return NWD(rhs, lhs % rhs);
+    else
+        return std::abs(lhs);
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    if(lhs != 0 && rhs != 0)
+    	return std::abs((lhs * rhs)/NWD(lhs, rhs));
```
piezol, line 13: I'm copying my comment from different PR:

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,12 @@
...
+        vec.push_back(step*i);
```
NeoBauer, line 10: There was a suggestion somewhere in our PR for Homework2, that it is better to add, than multiply that value. Possibility to get faster code ;).

ziobron:

PR #159, opened 2020-05-24
Fibonacci

dawidjozefczak:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,26 @@
 #pragma once
 
-int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+int fibonacci_iterative(int sequence){
+    int fibonacci[sequence];
```
piezol, line 4: Insecure code - static arrays should have size that is known at compile time.
If size of the array comes from a variable, the memory must be allocated dynamically.

Last but foremost - you only ever need at most two last values in the sequence, so you don't need an array at all.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,26 @@
...
+    return fibonacci[sequence];
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0){
+        return 0;
+    }
+    else if(sequence < 3){
```
piezol, line 17: For sequence == 2 we should go to the 'else' part - that way there are only two exceptions from the general rule, instead of three.

ziobron:

PR #161, opened 2020-05-24
Done list of the homework tasks from the Lesson Podstawy C++ 2 czesc 

mikailromer:
Define NWD and NWW function.
Define generateSequence function.
Define maxOfVector function.
Define addEven function.

### DIFF
--- a/module2/homework/addEven/addEven.hpp
+++ b/module2/homework/addEven/addEven.hpp
@@ -2,7 +2,10 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+    for (const auto& elem : numbers)
+    {
+        if((elem)%2 == 0) sum+= elem;
+    }
+    return sum;
 }
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -2,6 +2,14 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    if(count <= 0) return std::vector<int>{};
+    auto next_elem = step;
+    std::vector<int> sequence{};
+    sequence.reserve(count);
+    for(int i = 0; i < count; i++)
+    {
+        sequence.push_back(next_elem);
+        next_elem += step;
+    }
+    return sequence;
 }
--- a/module2/homework/maxOfVector/maxOfVector.hpp
+++ b/module2/homework/maxOfVector/maxOfVector.hpp
@@ -3,6 +3,10 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    auto max = vec[0];
+    for(const auto& elem : vec)
+    {
+        if(max < elem) max = elem;
+    }
+    return max;
 }
--- a/module2/homework/nwdNww/nwdNww.hpp
+++ b/module2/homework/nwdNww/nwdNww.hpp
@@ -55,7 +55,7 @@ int NWW(int lhs, int rhs) {
         divider = rhs;
     }
 
-    for(int i = dividend; i < 10000000; i++)
+    for(int i = dividend; ; i++)
     {
         if(i % dividend  == 0 && i % divider == 0)
         {
--- a/module2/homework/nwdNww/nwdNww.hpp
+++ b/module2/homework/nwdNww/nwdNww.hpp
@@ -37,6 +37,32 @@ int NWD(int lhs, int rhs) {
 }
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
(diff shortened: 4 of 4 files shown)


module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,10 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+    for (const auto& elem : numbers)
+    {
+        if((elem)%2 == 0) sum+= elem;
```
piezol, line 8:
```suggestion
        if(elem % 2 == 0) {
            sum += elem;
        }
```
Spacing

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,14 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    if(count <= 0) return std::vector<int>{};
```
piezol, line 5:
```suggestion
    if(count <= 0) {
        return std::vector<int>{};
    }
```

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,10 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    auto max = vec[0];
+    for(const auto& elem : vec)
+    {
+        if(max < elem) max = elem;
```
piezol, line 9:
```suggestion
        if(max < elem) {
            max = elem;
        }
```

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,68 @@
...
+    
+    if(rhs == 0 || lhs == 0)
+    {
+        if(rhs == 0 && lhs != 0) return lhs;
+        else if(rhs != 0 && lhs == 0) return rhs;
+        else return 0;
+    }
+
+    if(rhs < 0) rhs = abs(rhs);
+    if(lhs < 0) lhs = abs(lhs);
```
piezol, line 14:
```suggestion
    lhs = std::abs(lhs);
    rhs = std::abs(rhs);
```

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,68 @@
 #pragma once
```
piezol, line 2: You need to include a file that defines abs (or implement your own).

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,68 @@
...
 int NWD(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    int dividend, divider, result = 1;
+    
+    if(rhs == 0 || lhs == 0)
+    {
+        if(rhs == 0 && lhs != 0) return lhs;
+        else if(rhs != 0 && lhs == 0) return rhs;
+        else return 0;
```
piezol, line 10:
```suggestion
        if(rhs == 0 && lhs != 0) {
            return lhs;
        } else if(rhs != 0 && lhs == 0) {
            return rhs;
        }  
        return 0;

```
That's as formatting goes.
But i've seen even better solution: if either lhs or rhs is equal zero, return lhs + rhs.
No extra if statements.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,68 @@
...
 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    int dividend, divider, result = 1;
+
+    if(rhs == 0 || lhs == 0) return 0;
+
+    if(rhs < 0) rhs = abs(rhs);
+    if(lhs < 0) lhs = abs(lhs);
```
piezol, line 45:
```suggestion
    lhs = std::abs(lhs);
    rhs = std::abs(rhs);
```

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,68 @@
...
+    for(int i = dividend; ; i++)
+    {
+        if(i % dividend  == 0 && i % divider == 0)
+        {
+            result = i;
+            break;
+        }
+    }
+
+    return result;
```
piezol, line 67: Try to find more efficient algorithm for finding LCM. 
Your approach is very slow (it basically checks all the possibilities).

It is fine if your version doesn't return correct result for tricky inputs, but in your case there are inputs for which the algorithm would never end!
Try out 1e8-1, 1e8+1.
I overflows before it reaches LCM.

Even if the algorithm wouldn't require change:
loops without known number of iterations are usually better written using 'while'.
You don't need artificial variable to keep the result, you can just return it inside loop.

module2/homework/addEven/addEven.hpp
nauka-programowania-MA, line 8: space

ziobron:

PR #164, opened 2020-05-24
Calculate

KamilFinanseOrazProgramowanie:
Solution

### DIFF
--- a/module1/homework/calculate/calculate.hpp
+++ b/module1/homework/calculate/calculate.hpp
@@ -19,5 +19,5 @@ std::string calculate(const std::string& command, int first, int second) {
         }
         return std::to_string(first / second);
       }
-    return "Invalid data";
+  return "Invalid data";
 }
--- a/module1/homework/calculate/calculate.hpp
+++ b/module1/homework/calculate/calculate.hpp
@@ -26,6 +26,6 @@ std::string calculate(const std::string& command, int first, int second) {
     else {
         return "Invalid data";
     }
-    return 0;
+    return "";
 }
 }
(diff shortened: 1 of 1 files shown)


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,30 @@
 #include <string>
 
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
+    int result = 0;
+      
+    if (command == "add"){
+        result = first + second;
+        return std::to_string(result);
```
TomaszEichler, line 9: You can write it in shorter form : `return std::to_string(first + second)`

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,30 @@
...
+        if (second == 0){
+        return std::string("Division by Zero");
+        }
+        result = first / second;
+        return std::to_string(result);
+      }
+    else {
+        return "Invalid data";
+    }
     return "";
```
TomaszEichler, line 29: you just need here `return "Invalid data"`. `else` is unnecessary and last `return ""` also

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,30 @@
...
+        if (second == 0){
+        return std::string("Division by Zero");
+        }
+        result = first / second;
+        return std::to_string(result);
+      }
+    else {
+        return "Invalid data";
+    }
+    return 0;
```
MechaniQ1: This return is unnecessary, because function will always return value from if, elseif or else

ziobron:

PR #171, opened 2020-05-25
Homework2

marcindus:


module2/homework/addEven/addEven.hpp
```diff
@@ -1,8 +1,11 @@
...
 
-int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+int addEven(const std::vector<int>& numbers)
+{
+    std::vector<int> result;
+    std::copy_if(numbers.begin(), numbers.end(), std::back_inserter(result), [](int i) { return i % 2 == 0; });
+    return std::accumulate(result.begin(), result.end(), 0);
```
piezol, line 10: That's as clever as inefficient - why would we need that much memory to sum numbers in vector?

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -1,8 +1,63 @@
...
-    return {};
+int maxOfVector(const std::vector<int>& vec)
+{
+    BinaryTree bt;
+    std::shared_ptr<Node> current = nullptr;
+    for (auto i : vec)
+    {
+        current = bt.insert(current, i);
+    }
+    return findMax(current);
```
piezol, line 62: Overkill and an inefficient one- n*log(n) instead of O(n) that simplest implementation would require.
KISS

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,22 @@
...
+{
+    if (lhs == rhs)
+        return abs(lhs);
+    else if (lhs == 1 or rhs == 1)
+        return 1;
+    else if (lhs == 0)
+        return abs(rhs);
+    else if (rhs == 0)
+        return abs(lhs);
+    return abs(lhs) > abs(rhs) ? NWD(abs(lhs) - abs(rhs), abs(rhs)) : NWD(abs(rhs) - abs(lhs), abs(lhs));
```
piezol, line 14: Count the parentheses characters in this line

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,22 @@
...
 }
 
-int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+int NWW(int lhs, int rhs)
+{
+    if (lhs == 0 or rhs == 0)
+        return 0;
+    return abs(lhs * rhs) / NWD(lhs, rhs);
```
piezol, line 21: I'm copying my comment from different PR:

module2/homework/addEven/addEven.hpp
nauka-programowania-MA, line 10: Try without algorithms :)

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence
+8 XP delivery before 31.05

PR #172, opened 2020-05-25
everything works

mmichalm:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,43 @@
...
+    std::string result("0");
+	if(command == "add" || command == "subtract" || command == "multiply" || command == "divide")
+	{
+    char operation = command[0];
+	
+	switch (operation) 
+	{
+	case 'a':
+		result = std::to_string(first + second);
+		return result;
```
DamiXioR, line 14: in my opinion it's shorter to use directly return std::to_string(first + second);

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,43 @@
...
+		{
+			result = std::to_string(first / second);
+			return result;
+		}
+		else
+		{
+			
+			return "Division by 0"; 
+		}
+		break;
```
DamiXioR, line 35: i think this else is unnecessary, if case 'd' is last case label you do not need to use break statement (and it's just after last return - never execute)

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,43 @@
...
-    // TODO: Implement your solution here and return proper value
-    return "";
+    std::string result("0");
+	if(command == "add" || command == "subtract" || command == "multiply" || command == "divide")
+	{
+    char operation = command[0];
+	
+	switch (operation) 
+	{
+	case 'a':
```
piezol, line 12: That looks convenient 'for now' but requires major rework if list of commands ever changes.
Using switch doesn't really achieve anything here.

module1/homework/calculate/calculate.hpp
piezol: @DamiXioR is correct, return exits function body alltogether.

ziobron:

PR #173, opened 2020-05-25
adding works

danielzwierz:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,20 @@
...
+
+ std::string result = "Invalid data";
+
+  if(command == "add") result = std::to_string(first + second);
+  if(command == "subtract") result = std::to_string(first - second);
+  if(command == "multiply") result = std::to_string(first * second);
+  if(command == "divide") 
+  {
+	  if(second == 0) result = "Division by 0";  
+	  else result = std::to_string(first / second);
```
DamiXioR, line 15: i think it's better to use curly braces with if statement and skip else

danielzwierz: curly braces with if statement is optional. I am not sure how to skip else in this case, can you claryfy? 
```suggestion
	  result = "Division by 0";  
	  if(second <> 0) result = std::to_string(first / second);
```

ziobron:

PR #174, opened 2020-05-25
Fibonacci

mmichalm:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,43 @@
...
+	if(command == "add" || command == "subtract" || command == "multiply" || command == "divide")
+	{
+    char operation = command[0];
+	
+	switch (operation) 
+	{
+	case 'a':
+		result = std::to_string(first + second);
+		return result;
+		break;
```
kuchichan, line 15: You do not need break if you have a `return` statement.

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,43 @@
...
+		result = std::to_string(first * second);
+		return result;
+		break;
+	case 'd':
+		if (second != 0)
+		{
+			result = std::to_string(first / second);
+			return result;
+		}
+		else
```
kuchichan, line 30: You can check only `(second == 0)` and in that block `"return 'Division by 0"`. This way you can skip `else`

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,43 @@
 #include <string>
 
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+    std::string result("0");
+	if(command == "add" || command == "subtract" || command == "multiply" || command == "divide")
+	{
+    char operation = command[0];
```
kuchichan, line 8: I understand that you want to review `switch` statement, but here it is only make it too complicated. In production you would prefer:
``` c++
if ( command == "add") {
 ... 
} else if (command == "subtract") {
...
}
And  so on 

```

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,35 @@
 #pragma once
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
+    int a = 0;
+    int b = 1;
+    int var = 0;
+    if(sequence > 0)
+    {
```
kuchichan, line 8: If you want to do it that way - its fine, but i would say if(sequence <= 0) return (immediately). This way I would omit this for loop inside if :)

ziobron:

PR #178, opened 2020-05-25
[task] Solution for calculation method

embeddedMike:


module1/homework/calculate/calculate.hpp
```diff
@@ -3,5 +3,19 @@
...
+    if(command == "subtract") return std::to_string(first-second);
+    if(command == "multiply") return std::to_string(first*second);
+    if(command == "divide"){
+        if(second != 0)
+        {
+            return std::to_string(first/second);
+        }else
+        {
+            return "Division by 0";
+        }
```
niemarcin, line 17: I think it would be good to stick to one convention related to single-line if statement. For example, once you have

`if(command == "add") return std::to_string(first+second);`

then later why to use 
```
if(second != 0)
        {
            return std::to_string(first/second);
        }
```
?

------------------------------
Also, I would prefer space between if and parenthesis `if (something)` instead of `if(something)`. As well as space between ) and { in line 10. Also space between operands and operators would be nice: for example `(first + second)` instead of `(first+second)`.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,7 +2,24 @@
...
+	int firstValue = 1;
+	int result = 1;
+	if(sequence == 0)
+	{
+		return sequence;
+	}
+	if(sequence == 1)
+	{
+		return sequence;
+	}
```
niemarcin, line 15: These lines could be replaced with:
```
if (sequence == 0 || sequence == 1) {
    return sequence;
}
```

ziobron:

PR #181, opened 2020-05-26
Fibonacci works

SzymonGajewski:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,28 @@
...
+    int F1 = 1;
+    
+    if(sequence == 1 || sequence == 0){
+        Fnext = sequence;
+    }
+    else{
+        for(int i = 2; i <= sequence; ++i){        
+            Fnext = F0 + F1;
+            F0 = F1;
+            F1 = Fnext;
```
piezol, line 15: That loop does not have to be inside 'else'.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,28 @@
...
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 1 || sequence == 0){
+        return sequence;
+    }
+    else{
+        return fibonacci_recursive(sequence - 1) + fibonacci_recursive(sequence - 2);
+    }
```
DamiXioR, line 27: i think you can skip last else and just use return

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,28 @@
 #pragma once
 
-int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+int fibonacci_iterative(int sequence){
+    int Fnext{};
+    int F0{};
```
DamiXioR, line 5: in my opinion the better practice is to initialize variables with 0

SzymonGajewski: I thought that {} for int is same as = 0, isn't it? Or You mean just how looks better?

ziobron: It is the same and you can use them interchangeably. Which looks better is a personal matter :)

ziobron:

PR #183, opened 2020-05-26
All problems solved

kuchichan:
Feel free to review. My personal aim - keep it simple, and use things that we have already learned.

### DIFF
--- a/module2/homework/addEven/addEven.hpp
+++ b/module2/homework/addEven/addEven.hpp
@@ -4,7 +4,7 @@
 int addEven(const std::vector<int>& numbers) {
     auto sum = 0;
 
-    for (auto& number : numbers) {
+    for (const auto& number : numbers) {
 	if (number % 2 == 0) sum += number;
     }
     return sum;
--- a/module2/homework/addEven/addEven.hpp
+++ b/module2/homework/addEven/addEven.hpp
@@ -2,7 +2,10 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    auto sum = 0;
+
+    for (auto& number : numbers) {
+	if (number % 2 == 0) sum += number;
+    }
+    return sum;
 }
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -2,7 +2,7 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    std::vector<int> sequence { };
+    std::vector<int> sequence {};
 
     if (count < 0) return sequence;
 
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -6,6 +6,8 @@ std::vector<int> generateSequence(int count, int step) {
 
     if (count < 0) return sequence;
 
+    sequence.reserve(count);
+    
     for (std::size_t i = 0; i < count; ++i) {
 	sequence.push_back(step * (i + 1));
     }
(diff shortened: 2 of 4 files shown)


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> sequence { };
+
+    if (count < 0) return sequence;
+
```
piezol, line 8: The size of vector is known upfront, so we should allocate memory once instead of relying on push_back allocating memory whenever it's needed.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,11 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    auto max_number =  - std::numeric_limits<int>::max(); // the smallest number
```
piezol, line 6: Except that's wrong, since the smallest possible integer is NOT equal to max integer * (-1).

For a vector that consists only of a single number (-2147483648) your code will return (-2147483647), which isn't even in the vector.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,11 @@
...
+
+    for (auto& el : vec) {
+	if (el > max_number) max_number = el;
```
DamiXioR, line 9: you can use directly const auto reference

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> sequence { };
```
nauka-programowania-MA, line 5: remove space between {}

module2/homework/maxOfVector/maxOfVector.hpp
kuchichan: Fixed.

module2/homework/generateSequence/generateSequence.hpp
kuchichan: Fixed, but this would have huge impact? - I have red about "amortized constant time" which means that probably for bigger vectors, differences shall be negligible, what do you think?

module2/homework/generateSequence/generateSequence.hpp
kuchichan: Done.

module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,10 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    auto sum = 0;
+
+    for (auto& number : numbers) {
```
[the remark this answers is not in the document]
kuchichan: Fixed.

module2/homework/maxOfVector/maxOfVector.hpp
kuchichan: Thanks for that, fixed.

module2/homework/generateSequence/generateSequence.hpp
ziobron: I hope that after STL lesson you know how does it work :)
It's always better to tell the vector it's size upfront in a constructor (preferred) or at least in reserve().

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence
+8 XP delivery before 31.05

PR #193, opened 2020-05-27
Zadanie 1 - AddEven

WBigaj:


module2/homework/addEven/addEven.hpp
```diff
@@ -1,8 +1,15 @@
...
+#include <iostream>
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+    for (auto element : numbers) {
+        if(element % 2 == 0){
```
m-kop, line 8: formatting: missing spaces

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,9 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> result;
+    for(int i = 1;i <= count;i++){
+        result.push_back(step * i);
```
m-kop, line 7: consider adding step rather than multiply by i

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,23 @@
...
+#include <algorithm>
+#include <stdlib.h>
+#include <vector>
+
 
 int NWD(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    int first = std::max(abs(lhs), abs(rhs));
+    int second = std::min(abs(lhs), abs(rhs));
```
m-kop, line 9: you call abs() 4 times, consider calling it once on return value only.

module2/homework/addEven/addEven.hpp
```diff
@@ -1,8 +1,15 @@
 #pragma once
+#include <iostream>
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+    for (auto element : numbers) {
```
DamiXioR, line 7: i think you can use const auto reference

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,11 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    int max = std::numeric_limits<int>::min();
+    for (auto element : vec) {
```
DamiXioR, line 7: the same

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,9 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> result;
```
piezol, line 5: Consider reserving memory upfront if count > 0 (less memory allocations than depending on push_back)

module2/homework/nwdNww/nwdNww.hpp
piezol: There is no need for checking which number is greater and putting it on the 'left' side.
The algorithm is well thought out and even if you have greater number on the 'right' side, it will just do one iteration that switches the number, ie. your algorithm should work if you change your code to:
first = std::abs(lhs)
second = std::abs(rhs)

module2/homework/addEven/addEven.hpp
nauka-programowania-MA, line 7: since We learn const& please use it

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,9 @@
...
+    for(int i = 1;i <= count;i++){
```
nauka-programowania-MA, line 6: space

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA: agree. If you know exactly how many items will be in vector, please reserve memory.

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA: multiply by i is as good as adding step.

module2/homework/maxOfVector/maxOfVector.hpp
nauka-programowania-MA, line 7: const&

module2/homework/nwdNww/nwdNww.hpp
nauka-programowania-MA: just use 
first = std::abs(lhs)
second = std::abs(rhs)

module2/homework/addEven/addEven.hpp
WBigaj: Changed to const auto reference.

module2/homework/generateSequence/generateSequence.hpp
WBigaj: Adding is noticeable faster for larger sequence(like 1000). 
http://quick-bench.com/tY8T2mJcg3VAf8anNYWyZO0aino

module2/homework/maxOfVector/maxOfVector.hpp
WBigaj: the same :)

module2/homework/generateSequence/generateSequence.hpp
WBigaj: Applied

module2/homework/nwdNww/nwdNww.hpp
WBigaj: While m-kop has right that we can go down to use abs once on return. The double abs version that piezol and nauka-programowania-MA suggested seems to be faster for some reason.
http://quick-bench.com/BAf_MuD5gQUpiWo5kQ7XfKMl8zM

WBigaj: While m-kop has right that we can go down to calling abs just once in return, the double abs version that piezol and nauka-programowania-MA seems to be working faster for some reason.
http://quick-bench.com/BAf_MuD5gQUpiWo5kQ7XfKMl8zM

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,9 +2,14 @@
...
 std::vector<int> generateSequence(int count, int step) {
-    std::vector<int> result;
-    for(int i = 1;i <= count;i++){
-        result.push_back(step * i);
+    if (count <= 0){
+        return std::vector<int>{};
+    }
+    std::vector<int> result(count);
+    int sum = 0;
+    for (int i = 1;i <= count;i++){
```
piezol:
```suggestion
    for (int i = 1; i <= count; i++){
```

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence
+8 XP delivery before 31.05

PR #206, opened 2020-05-30
calculate working

shadston:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,28 @@
 #pragma once
 #include <string>
+using namespace std;
```
nauka-programowania-MA, line 3: please not using namespace. It could be misleading, when you will create bigger project and use extended library which can use similar names in function or sth else

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,28 @@
...
+        return to_string(first * second);
+    }
+
+    if(command == "divide") {
+        if(second == 0) {
+            return "Division by 0";
+        }
+        else {
+            return to_string(first / second);
+        }
```
nauka-programowania-MA, line 24: else not needed here

ziobron:

PR #207, opened 2020-05-30
fibonacci working

shadston:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,32 @@
...
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0) {
+        return 0;
+    }
+    if(sequence == 1) {
+        return 1;
+    }
```
Kasperice, line 30: You can notice some dependence between `sequence` and return value. Then you can simplify your solution

shadston: ok thank you

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,29 @@
...
```
nauka-programowania-MA, line 27: space after if

ziobron:

PR #220, opened 2020-05-31
Homework3

kontrybutor:
I tried to use C++20 standard but it seems that it's not supported on GitHub CI yet. Or I just couldn't run it properly. Changing CXX standard in CMakeLists.txt is not enough.

### DIFF
--- a/module3/homework/passwordCheck/validation.cpp
+++ b/module3/homework/passwordCheck/validation.cpp
@@ -5,55 +5,60 @@
 
 using EC = ErrorCode;
 
-std::string getErrorMessage(const ErrorCode& err)
+std::string getErrorMessage(const ErrorCode &err)
 {
     std::string message;
     std::map<ErrorCode, std::string> errorMap = {
-        { EC::Ok, "YourPasswordIsOK" },
-        { EC::PasswordNeedsAtLeastNineCharacters, "NeedsAtLeastNineCharacters" },
-        { EC::PasswordNeedsAtLeastOneNumber, "NeedsAtLeastOneNumber" },
-        { EC::PasswordNeedsAtLeastOneSpecialCharacter, "NeedsOneSpecialCharacter" },
-        { EC::PasswordNeedsAtLeastOneUppercaseLetter,
-            "NeedsAtLeastOneUppercaseLetter" },
-        { EC::PasswordsDoesNotMatch, "PasswordsDoesNotMatch" }
-    };
+        {EC::Ok, "YourPasswordIsOK"},
+        {EC::PasswordNeedsAtLeastNineCharacters, "NeedsAtLeastNineCharacters"},
+        {EC::PasswordNeedsAtLeastOneNumber, "NeedsAtLeastOneNumber"},
+        {EC::PasswordNeedsAtLeastOneSpecialCharacter, "NeedsOneSpecialCharacter"},
+        {EC::PasswordNeedsAtLeastOneUppercaseLetter,
+         "NeedsAtLeastOneUppercaseLetter"},
+        {EC::PasswordsDoesNotMatch, "PasswordsDoesNotMatch"}};
 
-    if (errorMap.contains(err)) {
+    if (errorMap.contains(err))
+    {
         message = errorMap[err];
     }
     return message;
 }
 
-bool doesPasswordsMatch(const std::string& passwd1,
-    const std::string& passwd2)
+bool doesPasswordsMatch(const std::string &passwd1,
+                        const std::string &passwd2)
 {
     return passwd1 == passwd2;
 }
 
-ErrorCode checkPasswordRules(const std::string& passwd)
+ErrorCode checkPasswordRules(const std::string &passwd)
 {
     EC err = EC::Ok;
 
-    if (std::any_of(passwd.begin(), passwd.end(), ::isupper)) {
+    if (std::any_of(passwd.begin(), passwd.end(), ::isupper))
+    {
         err = EC::PasswordNeedsAtLeastOneUppercaseLetter;
     }
     if (std::any_of(passwd.begin(), passwd.end(),
-            [](const auto c) { return !isalnum(c); })) {
+                    [](const auto c) { return !isalnum(c); }))
+    {
         err = EC::PasswordNeedsAtLeastOneSpecialCharacter;
     }
-    if (std::any_of(passwd.begin(), passwd.end(), ::isdigit)) {
+    if (std::any_of(passwd.begin(), passwd.end(), ::isdigit))
+    {
         err = EC::PasswordNeedsAtLeastOneNumber;
     }
-    if (passwd.length() < 9) {
+    if (passwd.length() < 9)
+    {
         err = EC::PasswordNeedsAtLeastNineCharacters;
     }
     return err;
 }
 
-ErrorCode checkPassword(const std::string& passwd1,
-    const std::string& passwd2)
+ErrorCode checkPassword(const std::string &passwd1,
+                        const std::string &passwd2)
 {
-    if (doesPasswordsMatch(passwd1, passwd2)) {
+    if (doesPasswordsMatch(passwd1, passwd2))
+    {
         return EC::Ok;
     }
     return EC::PasswordsDoesNotMatch;
--- a/module3/homework/passwordCheck/validation.cpp
+++ b/module3/homework/passwordCheck/validation.cpp
@@ -52,8 +52,8 @@ ErrorCode checkPasswordRules(const std::string &passwd)
     return EC::Ok;
 }
 
-ErrorCode checkPassword(const std::string &passwd1,
-                        const std::string &passwd2)
+ErrorCode checkPassword(const std::string& passwd1,
+                        const std::string& passwd2)
 {
     if (doesPasswordsMatch(passwd1, passwd2))
     {
--- a/module3/homework/passwordCheck/validation.cpp
+++ b/module3/homework/passwordCheck/validation.cpp
@@ -23,8 +23,8 @@ std::string getErrorMessage(const ErrorCode &err)
     return message;
 }
 
-bool doesPasswordsMatch(const std::string &passwd1,
-                        const std::string &passwd2)
+bool doesPasswordsMatch(const std::string& passwd1,
+                        const std::string& passwd2)
 {
     return passwd1 == passwd2;
 }
--- a/module3/homework/passwordCheck/validation.cpp
+++ b/module3/homework/passwordCheck/validation.cpp
@@ -2,7 +2,7 @@
 
 #include <algorithm>
 #include <map>
-
+#include <iostream>
 using EC = ErrorCode;
 
 std::string getErrorMessage(const ErrorCode &err)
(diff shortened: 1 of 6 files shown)


module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+
+    if (errorMap.find(err) != errorMap.end())
+    {
+        message = errorMap[err];
+    }
+    return message;
+}
+
+bool doesPasswordsMatch(const std::string &passwd1,
+                        const std::string &passwd2)
```
anraf1001, line 27: Connecting `&` with a type name makes it less possible to misunderstand that the `&` is a part of variables name.
```suggestion
bool doesPasswordsMatch(const std::string& passwd1,
                        const std::string& passwd2)
```

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+{
+    return passwd1 == passwd2;
+}
+
+ErrorCode checkPasswordRules(const std::string &passwd)
```
anraf1001, line 32:
```suggestion
ErrorCode checkPasswordRules(const std::string& passwd)
```

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+    if (!std::any_of(passwd.begin(), passwd.end(),
+                     [](const auto c) { return !isalnum(c); }))
+    {
+        return EC::PasswordNeedsAtLeastOneSpecialCharacter;
+    }
+    return EC::Ok;
+}
+
+ErrorCode checkPassword(const std::string &passwd1,
+                        const std::string &passwd2)
```
anraf1001, line 56:
```suggestion
ErrorCode checkPassword(const std::string& passwd1,
                        const std::string& passwd2)
```

module3/homework/passwordCheck/validation.hpp
```diff
@@ -1,2 +1,21 @@
...
+    PasswordsDoesNotMatch
+};
+
+std::string getErrorMessage(const ErrorCode &err);
+
+bool doesPasswordsMatch(const std::string &passwd1, const std::string &passwd2);
+
+ErrorCode checkPasswordRules(const std::string &passwd);
+
+ErrorCode checkPassword(const std::string &passwd1, const std::string &passwd2);
```
anraf1001, line 21: Enums are stored as ints, so I think that you don't need to use reference. Especially that you don't want to modify it.
```suggestion
std::string getErrorMessage(ErrorCode err);

bool doesPasswordsMatch(const std::string& passwd1, const std::string& passwd2);

ErrorCode checkPasswordRules(const std::string& passwd);

ErrorCode checkPassword(const std::string& passwd1, const std::string& passwd2);
```

module3/homework/vectorOfSharedPointers/vectorFunctions.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+std::vector<std::shared_ptr<int>> generate(int);
+
+void print(const std::vector<std::shared_ptr<int>> &);
+
+void add10(std::vector<std::shared_ptr<int>> &);
+
+void sub10(std::vector<std::shared_ptr<int>> &);
+
+void sub10(int *const);
```
anraf1001, line 14:
```suggestion
void print(const std::vector<std::shared_ptr<int>>&);

void add10(std::vector<std::shared_ptr<int>>&);

void sub10(std::vector<std::shared_ptr<int>>&);

void sub10(int* const);
```

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+void sub10(std::vector<std::shared_ptr<int>> &vec)
+{
+    for (const auto &elem : vec)
+    {
+        sub10(elem.get());
+    }
+}
+void sub10(int *const elem)
```
anraf1001, line 41:
```suggestion
void sub10(int* const elem)
```

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,47 @@
...
+    {
+        if (elem)
+        {
+            *elem += 10;
+        }
+    }
+}
+void sub10(std::vector<std::shared_ptr<int>> &vec)
+{
+    for (const auto &elem : vec)
```
anraf1001, line 36:
```suggestion
    for (const auto& elem : vec)
```

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,47 @@
...
+    {
+        if (elem)
+        {
+            *elem += 10;
+        }
+    }
+}
+void sub10(std::vector<std::shared_ptr<int>> &vec)
```
anraf1001, line 34:
```suggestion
void sub10(std::vector<std::shared_ptr<int>>& vec)
```

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,47 @@
...
+void print(const std::vector<std::shared_ptr<int>> &vec)
+{
+    for (const auto &elem : vec)
+    {
+        std::cout << *elem << '\n';
+    }
+}
+void add10(std::vector<std::shared_ptr<int>> &vec)
+{
+    for (const auto &elem : vec)
```
anraf1001, line 26:
```suggestion
void add10(std::vector<std::shared_ptr<int>>& vec)
{
    for (const auto& elem : vec)
```

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,47 @@
...
+    for (size_t i = 0; i < count; ++i)
+    {
+        vec.emplace_back(std::make_shared<int>(i));
+    }
+    return vec;
+}
+
+void print(const std::vector<std::shared_ptr<int>> &vec)
+{
+    for (const auto &elem : vec)
```
anraf1001, line 19:
```suggestion
void print(const std::vector<std::shared_ptr<int>>& vec)
{
    for (const auto& elem : vec)
```

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
 #include "validation.hpp"
-// TODO: Put implementations here
\ No newline at end of file
+
+#include <algorithm>
+#include <map>
+#include <iostream>
+using EC = ErrorCode;
+
+std::string getErrorMessage(const ErrorCode &err)
```
anraf1001, line 8: Enums are stored as ints, so I think that you don't need to use reference. Especially that you don't want to modify it.

module3/homework/passwordCheck/validation.hpp
kontrybutor: @nauka-programowania-MA Please comment if `enum class` should be pass by `const &` in this case. Usually there is no need to pass POD types via `&` but `enum class` is more complex type. Compiler probably optimize it so there won't be any difference in the machine code.

module3/homework/passwordCheck/validation.cpp
kontrybutor: Thanks. I think that one comment for the same function with the same insight is enough. If I correct this function signature only in header file I get linker errors so it's obvious I should correct the function definition in this source file as well.

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+{
+    std::string message;
+    std::map<ErrorCode, std::string> errorMap = {
```
piezol, line 11: That map could be static constexpr.

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+        {EC::Ok, "YourPasswordIsOK"},
+        {EC::PasswordNeedsAtLeastNineCharacters, "NeedsAtLeastNineCharacters"},
+        {EC::PasswordNeedsAtLeastOneNumber, "NeedsAtLeastOneNumber"},
+        {EC::PasswordNeedsAtLeastOneSpecialCharacter, "NeedsOneSpecialCharacter"},
+        {EC::PasswordNeedsAtLeastOneUppercaseLetter, "NeedsAtLeastOneUppercaseLetter"},
+        {EC::PasswordsDoesNotMatch, "PasswordsDoesNotMatch"}};
+
+    if (errorMap.find(err) != errorMap.end())
+    {
+        message = errorMap[err];
```
piezol, line 21: Find returns an iterator that points at the elenment you're looking for.
By using errorMap[err] instead of returned iterator, you're doing second lookup.

module3/homework/passwordCheck/validation.cpp
kontrybutor: I don't think so. `constexpr` is required to be of literal type. Please refer to [this](https://en.cppreference.com/w/cpp/named_req/LiteralType). `std::map` does not meet those requirements I think. If I am wrong, please provide me an working example of your suggestion :)

module3/homework/passwordCheck/validation.cpp
kontrybutor: Yes, you're right. My mistake.

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+    }
+    if (!std::any_of(passwd.begin(), passwd.end(), ::isupper))
+    {
+        return EC::PasswordNeedsAtLeastOneUppercaseLetter;
+    }
+    if (!std::any_of(passwd.begin(), passwd.end(), ::isdigit))
+    {
+        return EC::PasswordNeedsAtLeastOneNumber;
+    }
+    if (!std::any_of(passwd.begin(), passwd.end(),
```
Kasperice, line 47: I found another solution to this case. I can tell that it is based on `std::all_of` (it is a bit simpler), but your is also good :)

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
 #include "validation.hpp"
-// TODO: Put implementations here
\ No newline at end of file
+
+#include <algorithm>
+#include <map>
+#include <iostream>
```
nauka-programowania-MA, line 5: sort

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+using EC = ErrorCode;
```
nauka-programowania-MA, line 6: enter. separate using from icludes

module3/homework/passwordCheck/validation.cpp
nauka-programowania-MA, line 8: const& not needed here. Enum is usually represented as int

module3/homework/passwordCheck/validation.cpp
nauka-programowania-MA: std::string can't be constexpr. You need to use const char[]. But it will be more efficient when you try to construct this map as constexpr :)

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+}
+
+ErrorCode checkPassword(const std::string &passwd1,
+                        const std::string &passwd2)
+{
+    if (doesPasswordsMatch(passwd1, passwd2))
+    {
+        return EC::Ok;
+    }
+    return EC::PasswordsDoesNotMatch;
```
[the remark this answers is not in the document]
ziobron: Still there is a problem with this implementation. It's not correct. checkPasswordRules should be called when passwords match.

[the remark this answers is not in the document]
kontrybutor: Sure, I will fix it in next week. Unfortunately I have no time this week.

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,63 @@
...
+    if (!std::any_of(passwd.begin(), passwd.end(), ::isupper))
+    {
+        return EC::PasswordNeedsAtLeastOneUppercaseLetter;
+    }
+    if (!std::any_of(passwd.begin(), passwd.end(), ::isdigit))
+    {
+        return EC::PasswordNeedsAtLeastOneNumber;
+    }
+    if (!std::any_of(passwd.begin(), passwd.end(),
+                     [](const auto c) { return !isalnum(c); }))
```
WBigaj, line 48: You do not need lamba here. Isn't is isalnum functor. Even if not you can consider std::none_of and std::ispunct.

module3/homework/passwordCheck/validation.cpp
WBigaj: You even used another style in earlier function like const std::string& passwd.

ziobron:
{ "requirement_ids": [1998] }
5 XP for Podstawy#3 – vectorOfSharedPointers (1998)

No points for passwordCheck so far, because there is a bug. If you wish to deliver it please re-open this PR. I am closing it for now.

PR #229, opened 2020-05-31
Homework2

kasiasieradzka:


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -4,4 +4,19 @@
...
     // TODO: Implement me :)
     return {};
+
+    if (count <= 0){ 
+        return {};
+    }
+
+    auto firstValue = step;
+
+    std::vector<int> vecSequence;
```
piezol, line 14: You know the size of the vector upfront, so you can allocate memory once here.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -3,9 +3,44 @@
...
 
 int NWW(int lhs, int rhs) {
     // TODO: Implement me :)
     return -1;
+
+    if ((lhs == 0) || (rhs == 0)) {
+        return 0;
+    }
+
+    auto nww = (lhs * rhs) / NWD(lhs, rhs);
```
piezol, line 39: I'm copying my comment from different PR:

The simplest change that i saw in other solutions was to simply change the order of execution, ie.:
return (lhs / nwd(lhs, rhs)) * rhs.
the other possible solution is to cast lhs to long long before multiplication.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -5,4 +5,16 @@
 int maxOfVector(const std::vector<int>& vec) {
     // TODO: Implement me :)
     return {};
+
+    if (vec.empty() == true) {
+        return 0;
+    }
+
+    int maxValue = vec[0];
```
kuchichan, line 13: You can use `vec.front()` and this way you can also remove this `return 0` if empty. This can be misleading, because if vector has negative value and 0, the result will be the same.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -3,9 +3,44 @@
...
+
+    if (rhs == 0){
+        return lhs;
+    }
+    if (lhs == 0){
+        return rhs;
+    }
+
+    if (lhs < 0){
+        lhs = -1*lhs;
```
kuchichan, line 15: Pls check `std::abs`

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -3,9 +3,44 @@
...
+    }
+    if (rhs < 0){
+        rhs = -1*rhs;
+    }
```
Kasperice, line 19: You can use an absolute value function (`std::abs()`)

Kasperice: And one more thing here, you are opening brackets without space. And I see that you are using this space in other parts of code:
```suggestion
    if (lhs < 0) {
        lhs = -1 * lhs;
    }
    if (rhs < 0) {
        rhs = -1 * rhs;
    }
```
And if you don't want to use `std::abs()` please use whitespaces around operators.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -3,9 +3,44 @@
...
+
+    int divider = (lhs > rhs) ? lhs%rhs : rhs%lhs;
+
+    if (divider == 0)
+        return (lhs>rhs) ? rhs : lhs;
+    else
+        return (lhs > rhs) ? NWD(divider, rhs) : NWD (divider, lhs);
```
Kasperice, line 26: I saw that you are using brackets after `if` even if it has one line, so please add them here too.
```suggestion
    if (divider == 0) {
        return (lhs>rhs) ? rhs : lhs;
   } else {
        return (lhs > rhs) ? NWD(divider, rhs) : NWD (divider, lhs);
   }
```

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -4,4 +4,19 @@
 std::vector<int> generateSequence(int count, int step) {
     // TODO: Implement me :)
     return {};
```
ziobron, line 6: You should delete this line. It makes all test broken because you return always an empty vector. Please fix it :)

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -5,4 +5,16 @@
 int maxOfVector(const std::vector<int>& vec) {
     // TODO: Implement me :)
     return {};
```
ziobron, line 7: The same here. This return should be removed. It was added by me only to make this function compile.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -3,9 +3,44 @@
 int NWD(int lhs, int rhs) {
     // TODO: Implement me :)
     return -1;
```
ziobron, line 5: Remove it as well :)

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -3,9 +3,44 @@
...
+        return (lhs>rhs) ? rhs : lhs;
+    else
+        return (lhs > rhs) ? NWD(divider, rhs) : NWD (divider, lhs);
+
+    return divider;
 }
 
 int NWW(int lhs, int rhs) {
     // TODO: Implement me :)
     return -1;
```
ziobron, line 33: And the last important thing to remove.

kasiasieradzka:
@ziobron needed changes have made

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,15 @@
...
+        return 0;
+    }
+
+    int maxValue = vec[0];
+
+    for (const auto& el : vec) {
+        if (el > maxValue) {
+            maxValue = el;
+        }
+    }
```
ziobron, line 16: Now in tests, we can see that this implementation is not entirely correct. You are missing a return here and a random value is taken. Please add a proper return.
If you need help you can always take a look into other people implementations here: https://github.com/coders-school/kurs_cpp_podstawowy/pulls?q=is%3Apr+is%3Aclosed+sort%3Areactions-%2B1-desc

ziobron:
So far I can grant you points for NWD&NWW and generateSequence, because your implementation passed all tests. Working solutions were delivered before the deadline. And I am still waiting for a fixed maxOfVector. You can also deliver addEven. I see that it wasn't delivered yet.


+6 XP NwwNwd
+5 XP generateSequence
+4 XP for delivering 2 tasks before 31.05

{ "requirement_ids": [1119, 1633, 1119, 1635], "points": [2, 6, 2, 5] }

PR #232, opened 2020-05-31
Homework2

majkel84:


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    if (count <= 0) {
+        return {};
+    }
+    std::vector<int> vec {step};
```
Kasperice, line 8: Are you reserving space for vector of size = step? Or should it be equal to count?

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
...
+    for (auto it = 1; it < count; it++) {
+        vec.emplace_back(vec[it - 1]+ step);
```
Kasperice, line 10:
```suggestion
        vec.emplace_back(vec[it - 1] + step);
```

module2/homework/generateSequence/generateSequence.hpp
majkel84: i'm creating vector with just one value, don't need to reserve any adittional space in this step

piezol: "Don't need to" - but it's beneficial to allocate memory just once instead of multiple times.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,34 @@
...
-    return -1;
+    if (rhs == 0) {
+        return lhs;
+    }
+    if (lhs == 0) {
+        return rhs;
+    }
+    if (lhs == rhs) {
+        return lhs;
+    }
```
piezol, line 12: That's wrong.
If they are both negative, you would return negative number, ie. for (lhs = -3, rhs = -3) your function returns -3 instead of 3.
You can avoid having such errors by generalizing algorithm when it's possible.
Instead of using abs in lines 13 and 14 you could use abs at the very beginning of the function. Less cases to worry about in later code = less possibilities of having an error.

module2/homework/generateSequence/generateSequence.hpp
majkel84: so after initialising i schuld ad vec.reserve(count), tnks for an advice

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,34 @@
...
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    if (lhs == 0 || rhs ==0) {
+        return 0;
+    }
+    if (abs(lhs) == abs(rhs)) {
+        return abs(lhs);
+    }
+    int result = abs(lhs * rhs) / (NWD(lhs, rhs));
```
piezol, line 32: I'm copying my comment from different PR:

The simplest change that i saw in other solutions was to simply change the order of execution, ie.:
return (lhs / nwd(lhs, rhs)) * rhs.
the other possible solution is to cast lhs to long long before multiplication.

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence
+8 XP delivery before 31.05

PR #234, opened 2020-05-31
Podstawy C++ #2

WojtekMs:
Authors: @mmichalm, @WojtekMs


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,25 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    
+    std::vector<int> vec;
+    if(count > 0)
```
kuchichan, line 7: It will be simplier to check `count == 0` and return vector immediately. This way you will not store the entire loop in if.

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,25 @@
...
+    {
+    vec.resize(count);
+    for(size_t i = 0; i < vec.size(); ++i)
+    {   if(i == 0)
+        {
+            vec[i] = step;
+        }
+        if(i != 0) 
+        {
+        vec[i] = vec[i-1] + step;
```
kuchichan, line 17: Pls leverage `push_back()`

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,25 @@
...
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    
+    std::vector<int> vec;
+    if(count > 0)
+    {
+    vec.resize(count);
+    for(size_t i = 0; i < vec.size(); ++i)
+    {   if(i == 0)
```
kuchichan, line 11: If you change the way of "computing" element of vec (by adding 1) you can get rid of one `if`

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA: agree

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,11 @@
 #include <limits>
```
nauka-programowania-MA, line 3: remove if not use

module2/homework/generateSequence/generateSequence.hpp
WojtekMs: I think checking count == 0 does not cover all the possibilities, shouldn't it be count <= 0?

module2/homework/generateSequence/generateSequence.hpp
WojtekMs: Why? Is it not better to allocate memory to vector only once if we know the desired size from the beginning?

module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,11 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+    for(const auto & elem : numbers) {
```
Vicx95, line 6:
```suggestion
    for(const auto& elem : numbers) {
```

module2/homework/generateSequence/generateSequence.hpp
ziobron: Yes, indeed :)

module2/homework/generateSequence/generateSequence.hpp
ziobron: It is. If you know the size upfront you can take advantage of it.

Code formatting is bad here. This line should be indented.

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence
+8 XP delivery before 31.05
+3 XP cooperation

{
  "requirement_ids": [1119, 1631, 1119, 1633, 1119, 1634, 1119, 1635, 1636],
  "points": [2, 4, 2, 6, 2, 5, 2, 5, 3],
  "users": ["mmichalm", "WojtekMs"]
}

PR #238, opened 2020-05-31
Homework2

TomaszEichler:


module2/homework/addEven/addEven.hpp
```diff
@@ -4,5 +4,12 @@
 int addEven(const std::vector<int>& numbers) {
     // TODO: Your implementation goes here
     // Below return is only to make this function compile now
-    return -1;
+    int sum {};
+
+    for(const auto& element : numbers){
+        if(!(element % 2)){
```
piezol, line 10: Implicit casts are usually a code smell.
Comparison to 0 is much clearer
```suggestion
        if(element % 2 == 0) {
```

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,16 @@
...
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> myVector{};
+
+    if(count <= 0){
+        return myVector;
+    }
+
```
piezol, line 10: Try to preallocate memory for the whole vector upfront - you know the size of the vector at the end.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,30 @@
...
+        return rhs;
+    }
+    else if(rhs == 0){
+        return lhs;
+    }
+
+    int temp;
+    while(rhs != 0){
+        temp = std::abs(rhs);
+        rhs = std::abs(lhs) % temp;
```
piezol, line 15: You have two calls two abs in every iteration.
The algorithm will work correctly if you call abs before the loop and always work on positive values.

TomaszEichler: Thank you, great idea :)

module2/homework/generateSequence/generateSequence.hpp
TomaszEichler: Yeah, you're right I should do that after `if` condition. Good advice, thank you :)

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence
+8 XP delivery before 31.05

Issue #233, opened 2020-05-31
Semicolons in enum definition


kuchichan:
In presentation (enum chapter) there is a small mistake regarding to enum definition (semicolons instead of commas).
``` c++
enum ErrorCode {
    lack_of_water;
    too_much_load;
    bearing_problem;
    block_of_pump;
};

// or better ↓

enum class ErrorCode {
    lack_of_water;
    too_much_load;
    bearing_problem;
    block_of_pump;
};
```
IMHO should be:
``` c++
enum ErrorCode {
    lack_of_water,
    too_much_load,
    bearing_problem,
    block_of_pump,
};

// or better ↓

enum class ErrorCode {
    lack_of_water,
    too_much_load,
    bearing_problem,
    block_of_pump,
};
```

PR #239, opened 2020-05-31
Changes in presentation_enums.md

niemarcin:
In enum definitions semicolons should be changed to commas

### DIFF
--- a/module3/presentation_enums.md
+++ b/module3/presentation_enums.md
@@ -35,19 +35,19 @@ ___
 
 ```cpp
 enum ErrorCode {
-    lack_of_water;
-    too_much_load;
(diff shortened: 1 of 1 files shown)


module3/presentation_enums.md
```diff
@@ -135,7 +135,7 @@ Aby uniknąć konfliktu nazw stosujemy `enum class`.
 enum class Color {
     RED,
     GREEN,
-    BLUE,
+    BLUE
```
ziobron, line 138: it's not a bug, you can have a comma after the last element of enum. However, it is better and more consistent when there is no comma.

ziobron:
Closes #233

PR #245, opened 2020-06-02
Homework Podstawy C++ #3

anraf1001:
Authors: @Vicx95, @anraf1001

### DIFF
--- a/module3/homework/passwordCheck/test.cpp
+++ b/module3/homework/passwordCheck/test.cpp
@@ -4,6 +4,9 @@
 std::string EMPTY_PASSWORD = "";
 std::string PROPER_PASSWORD = "abcABC123!@#";
 std::string TOO_SHORT_PASSWORD = "12345678";
+std::string PASSWORD_WITHOUT_DIGITS = "abcdefghijk";
+std::string PASSWORD_WITHOUT_SPECIAL_CHAR = "123abcdefghijk";
+std::string PASSWORD_WITHOUT_UPPERCASE_LETTER = "123abcdefghi?!";
 auto MIN_ALLOWED_ERROR_CODE = ErrorCode::Ok;
 auto MAX_ALLOWED_ERROR_CODE = ErrorCode::PasswordNeedsAtLeastOneUppercaseLetter;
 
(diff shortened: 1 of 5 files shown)


module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,45 @@
+#include "vectorFunctions.hpp"
+
+#include <algorithm>
+#include <functional>
+#include <iostream>
+
+std::vector<std::shared_ptr<int>> generate(int count) {
+    std::vector<std::shared_ptr<int>> result;
```
krzyout, line 8: You know the size of vector, you can use vector.reserve().

KrzysztofJ85, line 8: You know the size of vector (count), you can use reserve().

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,58 @@
...
+    return ErrorCode::Ok;
+}
+
+ErrorCode checkPassword(const std::string& firstPwd, const std::string& secondPwd) {
+    if (!doesPasswordsMatch(firstPwd, secondPwd)) {
+        return ErrorCode::PasswordsDoesNotMatch;
+    }
+
+    return checkPasswordRules(firstPwd);
+}
```
m-kop, line 58: newline

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,58 @@
 #include "validation.hpp"
-// TODO: Put implementations here
\ No newline at end of file
+
+std::string getErrorMessage(ErrorCode codeToShow) {
+    switch (codeToShow) {
+    case ErrorCode::Ok:
+        return "OK";
+        break;
```
kuchichan, line 7: I think `break` is not needed if you are using return one line above.

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,52 @@
...
+    case ErrorCode::PasswordNeedsAtLeastOneNumber:
+        return "Password needs at least one number";
+    case ErrorCode::PasswordNeedsAtLeastOneSpecialCharacter:
+        return "Password needs at least one special character";
+    case ErrorCode::PasswordNeedsAtLeastOneUppercaseLetter:
+        return "Password needs at least one uppercase letter";
+    case ErrorCode::PasswordsDoesNotMatch:
+        return "Passwords don't match";
+    default:
+        return "Unknown error code";
```
DamiXioR, line 18: use the curly braces even on one line statements

module3/homework/passwordCheck/validation.hpp
```diff
@@ -1,2 +1,22 @@
...
+    PasswordNeedsAtLeastOneSpecialCharacter,
+    PasswordNeedsAtLeastOneUppercaseLetter,
+    PasswordsDoesNotMatch
+};
+
+std::string getErrorMessage(ErrorCode codeToShow);
+
+bool doesPasswordsMatch(const std::string& firstPwd, const std::string& secondPwd);
+
+ErrorCode checkPasswordRules(const std::string& password);
```
DamiXioR, line 20: I think you can call this variable like before: firstPwd

module3/homework/passwordCheck/validation.cpp
anraf1001: Do you want me to use curly brackets inside every case in `switch case`?

module3/homework/passwordCheck/validation.hpp
anraf1001: `first` can suggest that there is a second one, but there is only one parameter :/

kuchichan: I agree with @anraf1001.

module3/homework/passwordCheck/validation.cpp
piezol: I don't think that the're really needed here.
That's simple case-return-case-return-case-return.
It's more readable without braces.

anraf1001: I don't think that you really need any braces inside cases even when they are longer...

module3/homework/passwordCheck/test.cpp
```diff
@@ -26,15 +29,23 @@ TEST(checkPasswordRulesTests, returnsValuesBetweenOkAndPasswordNeedsAtLeastOneUp
     EXPECT_LE(checkPasswordRules(TOO_SHORT_PASSWORD), MAX_ALLOWED_ERROR_CODE);
 }
 
+TEST(checkPasswordRulesTests, returnExactErrorMessage) {
```
nauka-programowania-MA, line 32: extract

module3/homework/passwordCheck/validation.cpp
nauka-programowania-MA: you have break; to tell compiler when should end. Bracket not needed here.

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,52 @@
...
+ErrorCode checkPasswordRules(const std::string& password) {
+    if (password.length() < 9) {
+        return ErrorCode::PasswordNeedsAtLeastNineCharacters;
+    }
+
+    if (!std::any_of(password.begin(), password.end(), isdigit)) {
+        return ErrorCode::PasswordNeedsAtLeastOneNumber;
+    }
+
+    if (std::none_of(password.begin(), password.end(), [](char sign) { return isprint(sign) && !isalnum(sign); })) {
```
nauka-programowania-MA, line 35: too long

module3/homework/passwordCheck/validation.hpp
```diff
@@ -1,2 +1,22 @@
-// TODO: I'm empty :) Put enum and function headers here.
-// Don't forget the header guard
\ No newline at end of file
+#pragma once
+
+#include <algorithm>
+#include <array>
```
nauka-programowania-MA, line 4: remove

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
nauka-programowania-MA, line 8: please always reserve memory, when you get knowledge, how big should it be.

module3/homework/passwordCheck/validation.hpp
```diff
@@ -1,2 +1,22 @@
...
+#include <iostream>
```
NG90, line 5: Those includes are not needed in header file.
Only one that would be required here is <string> library.

anraf1001: It's been already mentioned above

PR #249, opened 2020-06-02
Palindrome

WBigaj:


Palindrome/Palindrome.cpp
```diff
@@ -0,0 +1,20 @@
+#include <iostream>
+
+bool isPalindrome(const std::string &text){
```
danielzwierz, line 3: spaces, check your code format

Palindrome/Palindrome.cpp
```diff
@@ -0,0 +1,20 @@
...
+    bool result = true;
+    for (long unsigned int i = 0;i< text.length() / 2;i++){
```
danielzwierz, line 5: spaces

module2/homework/addEven/addEven.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+#include <iostream>
+#include <vector>
+
+int addEven(const std::vector<int>& numbers) {
+    int sum = 0;
+    for (const auto& element : numbers) {
+        if (element % 2 == 0){
```
danielzwierz, line 8: space

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+#include <vector>
+
+std::vector<int> generateSequence(int count, int step) {
+    if (count <= 0){
+        return std::vector<int>{};
```
danielzwierz, line 6: return {}

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -0,0 +1,15 @@
...
+    }
+    std::vector<int> result(count);
+    int sum = 0;
+    for (int i = 1;i <= count;i++){
+        sum += step;
+        result[i - 1] = sum;
```
danielzwierz, line 12: you are using i only in result[i - 1]. try to start loop with i=0, so then you dont need to make this sub. try to use i * step instead of sum, you can save 2 line of code and one variable.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include <vector>
+#include <limits>
```
danielzwierz, line 3: sort, check QA lesson

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -0,0 +1,13 @@
...
+
+int maxOfVector(const std::vector<int>& vec) {
+    int max = std::numeric_limits<int>::min();
+    for (const auto& element : vec) {
+        if (element > max){
```
danielzwierz, line 8: space

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include <algorithm>
+#include <stdlib.h>
+#include <vector>
+
+
+int NWD(int lhs, int rhs) {
+    int first = abs(lhs);
+    int second = abs(rhs);
+    while (second > 0){
```
danielzwierz, line 10: you use abs for second, try while(second), you can also try to use abs once on resoults intead of twice for first and secound

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -0,0 +1,23 @@
...
+        int tmp = first % second;
+        first = second;
+        second = tmp;
+    }
+    return first;
+}
+
+int NWW(int lhs, int rhs) {
+    if (lhs == 0 && rhs == 0){
```
danielzwierz, line 19: space, or

Palindrome/Palindrome.cpp
Vicx95, line 3:
```suggestion
bool isPalindrome(const std::string& text){
```

.github/workflows/main.yml
```diff
@@ -3,35 +3,66 @@ name: C++ with CMake CI
 on: [push, pull_request]
```
m-kop, line 3: do not commit files from .xxx directories

WBigaj:
OMG! I just wanted to send palindrome on palindrome branch!But I had send my homework2 also. I will try to fix after work.

module2/homework/generateSequence/generateSequence.hpp
WBigaj: Thanks for teaching me that I can use return {} to achieve the same result,but I think I will stay my current more transparent version.

module2/homework/generateSequence/generateSequence.hpp
WBigaj: You are right loop iterator. I changed that. I tried version with "i * step", but additive version is faster.
http://quick-bench.com/tY8T2mJcg3VAf8anNYWyZO0aino

module2/homework/nwdNww/nwdNww.hpp
WBigaj: I tried to go down to calling abs once,but for some reason version with two abs is faster.
http://quick-bench.com/BAf_MuD5gQUpiWo5kQ7XfKMl8zM

PR #251, opened 2020-06-02
Homework3

Marcinati:
Authors: @Kasperice @Marcinati


module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,62 @@
...
+    if (password.size() < 9) {
+        return ErrorCode::PasswordNeedsAtLeastNineCharacters;
+    } else if (!std::any_of(password.begin(), password.end(), ::isdigit)) {
+        return ErrorCode::PasswordNeedsAtLeastOneNumber;
+    } else if (!std::any_of(password.begin(), password.end(), ::ispunct)) {
+        return ErrorCode::PasswordNeedsAtLeastOneSpecialCharacter;
+    } else if (!std::any_of(password.begin(), password.end(), ::isupper)) {
+        return ErrorCode::PasswordNeedsAtLeastOneUppercaseLetter;
+    } else {
+        return ErrorCode::Ok;
```
danielzwierz, line 52: I am not an expert but i think  you should start else in new line

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,62 @@
...
+    }
+}
+
+ErrorCode checkPassword(const std::string& password, const std::string& repeatedPassword) {
+    if (doesPasswordsMatch(password, repeatedPassword)) {
+        return checkPasswordRules(password);
+    } else {
```
danielzwierz, line 59: you dont need this else

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,62 @@
...
-// TODO: Put implementations here
\ No newline at end of file
+#include <algorithm>
+#include <string>
+
+std::string getErrorMessage(ErrorCode errorNumber) {
+    switch (errorNumber) {
+        case ErrorCode::Ok: {
+            return "OK";
+            break;
```
kuchichan, line 9: No need to break if you `return` - function will exit in this place anyway

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,39 @@
+#include "vectorFunctions.hpp"
+
+std::vector<std::shared_ptr<int>> generate(int count) {
+    std::vector<std::shared_ptr<int>> result_vector;
+    result_vector.reserve(count);
+    for (size_t i = 0; i < count; i++) {
+        result_vector.push_back(std::make_shared<int>(i));
```
danielzwierz, line 7: it is better to use emplace_back

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,39 @@
...
+    }
+    return result_vector;
+}
+
+void print(const std::vector<std::shared_ptr<int>>& vec) {
+    for (const std::shared_ptr<int> ptr : vec) {
```
danielzwierz, line 13: you can use here const auto&

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,39 @@
...
+        std::cout << *ptr << "\t";
+    }
+    std::cout << '\n';
+}
+
+void add10(std::vector<std::shared_ptr<int>>& vec) {
+    for (const std::shared_ptr<int> ptr : vec) {
```
danielzwierz, line 20: you can use here const auto&

kuchichan, line 20: IMHO `const auto& ptr` will be viable too, but let's wait for some "real" reviewer :)

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+void sub10(int* const ptr) {
+    if (ptr) {
+        *ptr -= 10;
+    }
+}
+
+void sub10(std::vector<std::shared_ptr<int>>& vec) {
+    for (std::shared_ptr<int> ptr : vec) {
```
kuchichan, line 34: The same as above.

module3/homework/passwordCheck/validation.cpp
danielzwierz: It was discussed in previous reviews. Break is a good practice, but also somebody said that it can cause some compile errors. I used breaks and it was ok.

module3/homework/passwordCheck/validation.cpp
kuchichan: I think that it was agreed on previous lesson that the only approved style is:
`{ BasedOnStyle: Chromium, UseTab: Never, IndentWidth: 4, TabWidth: 4, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4 }` so it shouldn't be the matter of opinion but rather autoformatters' job.

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
KrzysztofJ85: > you can use here const auto&

Does using `const auto&` have any advantages besides looking better for someone?

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
KrzysztofJ85: > it is better to use emplace_back

Can you elaborate why it this case?

module3/homework/passwordCheck/validation.cpp
KrzysztofJ85: As I remember, @nauka-programowania-MA said that it is good practice to put break; anyway.

module3/homework/passwordCheck/validation.cpp
Kasperice: We formatted it using this style. For me starting else in same line as ending bracket is better because code is more compact. Especially if it is not complicated (like here)

module3/homework/passwordCheck/validation.cpp
Kasperice: Yes, I found in previous review that it is a good practice that is why we used that

kuchichan: Ok, so maybe I shall catch up ;)  Sorry for that.

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
DamiXioR: true but without auto It's easier to understand what he did

module3/homework/passwordCheck/validation.cpp
DamiXioR: yes and always even you don't have return, don't put break in last case or default, it's redundant

module3/homework/passwordCheck/validation.cpp
DamiXioR: you can also use lamda expression, with if statement and ANSI relationship

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
DamiXioR: I think id doesn't look better. Is just easier to use.

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
DamiXioR: because shared pointer is a class and when You use emplace_back the element is constructed in place

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,60 @@
...
+ErrorCode checkPasswordRules(const std::string& password) {
+    if (password.size() < 9) {
+        return ErrorCode::PasswordNeedsAtLeastNineCharacters;
+    } else if (!std::any_of(password.begin(), password.end(), ::isdigit)) {
+        return ErrorCode::PasswordNeedsAtLeastOneNumber;
+    } else if (!std::any_of(password.begin(), password.end(), ::ispunct)) {
+        return ErrorCode::PasswordNeedsAtLeastOneSpecialCharacter;
+    } else if (!std::any_of(password.begin(), password.end(), ::isupper)) {
+        return ErrorCode::PasswordNeedsAtLeastOneUppercaseLetter;
+    } else {
```
nauka-programowania-MA, line 50: else not needed

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
nauka-programowania-MA: Basically. When you want construct something in container (not, copy) it's good practise to use `emplace_back`.
Because when you use puhs_back. Compiler will create temporary object, which will be move to container memory. emplace_back will construct this object directly in memory of container without create temporary object.

Kasperice:
@ziobron pewnie już namierzyłeś tego PR, ale tu nam nie naliczyło punktów

PR #254, opened 2020-06-02
Homework3

NG90:


module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+void sub10(int* const element) {
+    if (element) {
+        *element -= 10;
+    }
+}
+
+void sub10(const std::vector<std::shared_ptr<int>>& sharedPointersVector) {
+    for (auto& el : sharedPointersVector) {
```
Vicx95, line 41:
```suggestion
    for (const auto& el : sharedPointersVector) {
```

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+void print(const std::vector<std::shared_ptr<int>>& sharedPointersVector) {
+    for (const auto& el : sharedPointersVector) {
+        std::cout << *el << " | ";
+    }
+    std::cout << '\n';
+}
+
+void add10(const std::vector<std::shared_ptr<int>>& sharedPointersVector) {
+    for (auto& el : sharedPointersVector) {
```
Vicx95, line 27:
```suggestion
    for (const auto& el : sharedPointersVector) {
```

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <iostream>
+
+std::vector<std::shared_ptr<int>> generate(int count) {
+    std::vector<std::shared_ptr<int>> resultVector{};
+
+    if (count < 0) {
+        return resultVector;
+    }
+
+    resultVector.reserve(count);
```
Vicx95, line 12: You can specify a size of vector in constructor :)
```suggestion
    std::vector<std::shared_ptr<int>> resultVector(count);

    if (count < 0) {
        return resultVector;
    }

```

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
NG90: done, thanks

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
NG90: done, thanks

module3/homework/vectorOfSharedPointers/vectorFunctions.cpp
NG90: Yes, but what about situation, when count is negative value? I do not think such value should be used in the vector constructor, that's why I do such check first.

Vicx95: you are right :) , my fault :)

NG90: No problem, thanks for review :)

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,62 @@
...
+    case ErrorCode::PasswordNeedsAtLeastOneNumber:
+        return "Password needs at least one number";
+    case ErrorCode::PasswordNeedsAtLeastOneSpecialCharacter:
+        return "Password needs at least one special character";
+    case ErrorCode::PasswordNeedsAtLeastOneUppercaseLetter:
+        return "Password needs at least one uppercase letter";
+    case ErrorCode::PasswordsDoesNotMatch:
+        return "Passwords does not match";
+    default:
+        return "Unknown error";
```
DamiXioR, line 25: I think it's better when every case has curly braces even it's has one line code

module3/homework/passwordCheck/validation.cpp
```diff
@@ -1,2 +1,62 @@
...
+    }
+}
+
+bool doesPasswordsMatch(const std::string& basePassword, const std::string& repeatedPassword) {
+    return basePassword == repeatedPassword;
+}
+
+ErrorCode checkPasswordRules(const std::string& password) {
```
DamiXioR, line 33: nice implementation, You can use also lambda expression, conditional expression and  relationship ANSI

module3/homework/passwordCheck/validation.cpp
NG90: Well, case works a little bit different than other statements (if, else or loops) and even several lines of code will be handled properly without curly braces.
I think this is matter of style agreement that we should define for the project.

module3/homework/passwordCheck/validation.cpp
NG90: Thanks. I used lambdas by the way :)
Can you share more info about relationship ANSI? I do not think I get your point.

DamiXioR: I mean that ASCII code 97 is interpreted as the character 'a'. Etc. If char is bigger than 64 and lesser than 91 char is a capital letter

DamiXioR: ASCII no ANSI my bad

NG90: Yes, sure - this was my first idea. However, since standard library gives possibility to use built-in functions why don't use them :)

DamiXioR: true!

module3/homework/passwordCheck/test.cpp
```diff
@@ -26,18 +31,37 @@ TEST(checkPasswordRulesTests, returnsValuesBetweenOkAndPasswordNeedsAtLeastOneUp
...
+              "Password needs at least nine characters");
+    EXPECT_EQ(getErrorMessage(ErrorCode::PasswordNeedsAtLeastOneNumber),
+              "Password needs at least one number");
+    EXPECT_EQ(getErrorMessage(ErrorCode::PasswordNeedsAtLeastOneSpecialCharacter),
+              "Password needs at least one special character");
+    EXPECT_EQ(getErrorMessage(ErrorCode::PasswordNeedsAtLeastOneUppercaseLetter),
+              "Password needs at least one uppercase letter");
+    EXPECT_EQ(getErrorMessage(ErrorCode::PasswordsDoesNotMatch), "Passwords does not match");
+    EXPECT_EQ(getErrorMessage(static_cast<ErrorCode>(500)), "Unknown error");
 }
```
nauka-programowania-MA, line 67: new line

PR #257, opened 2020-06-03
sorted headers of maxOfVector solution

Marcinati:


Marcinati:
> I think You should PR all solutions by branch homework2

It's not my homework. ;)

PR #258, opened 2020-06-03
My version of palindrome from intro module

WBigaj:
So I managed to create pull request with palindrome on new branch with only palindrome. Thanks for DamiXioR,danielzwierz, Vicx95 and m-kop for feedback for both palindrome and homework from module2.

### DIFF
--- a/Palindrome/Palindrome.cpp
+++ b/Palindrome/Palindrome.cpp
@@ -1,0 +1,20 @@
+#include <iostream>
+
+bool isPalindrome (const std::string& text) {
+    bool result = true;
+    for (long unsigned int i = 0;i< text.length() / 2;i++) {
+        if (text[i] != text[text.length() - i - 1]) {
+            result = false;
+            break;
+        }
+    }
+    return result;
+}
+
+int main() {
+    std::cout << isPalindrome("Ala ma kota") << "\n";
+    std::cout << isPalindrome("ala") << "\n"; 
+    std::cout << isPalindrome("kajak") << "\n";
+    std::cout << isPalindrome("neveroddoreven") << "\n";
+    std::cout << isPalindrome("abacada") << "\n";
+}


Palindrome/Palindrome.cpp
```diff
@@ -0,0 +1,20 @@
...
+    return result;
+}
+
+int main() {
+    std::cout << isPalindrome("Ala ma kota") << "\n";
+    std::cout << isPalindrome("ala") << "\n"; 
+    std::cout << isPalindrome("kajak") << "\n";
+    std::cout << isPalindrome("neveroddoreven") << "\n";
+    std::cout << isPalindrome("abacada") << "\n";
+}
```
kyrtaq, line 20: always add a new line

Palindrome/Palindrome.cpp
```diff
@@ -0,0 +1,20 @@
+#include <iostream>
+
+bool isPalindrome (const std::string& text) {
+    bool result = true;
+    for (long unsigned int i = 0;i< text.length() / 2;i++) {
```
kyrtaq, line 5: better to use size_t instead of long unsigned int

Palindrome/Palindrome.cpp
```diff
@@ -0,0 +1,20 @@
+#include <iostream>
+
```
kyrtaq, line 2: if you want to use strings you should #include <string>

Palindrome/Palindrome.cpp
```diff
@@ -0,0 +1,20 @@
...
+    }
+    return result;
+}
+
+int main() {
+    std::cout << isPalindrome("Ala ma kota") << "\n";
+    std::cout << isPalindrome("ala") << "\n"; 
+    std::cout << isPalindrome("kajak") << "\n";
+    std::cout << isPalindrome("neveroddoreven") << "\n";
+    std::cout << isPalindrome("abacada") << "\n";
```
kyrtaq, line 19: you dont have to use "\n" you can use it as a one character '\n' it works the same but doesnt generate '\0'

Palindrome/Palindrome.cpp
kyrtaq, line 20: you specified that main returns int, but you didnt return it, i would suggest to put return 0; or change it to void main

piezol: @kyrtaq void main is C, not C++ :)

ziobron:
Thanks @WBigaj. There were no additional points for this task, but thank you for your implementation. Let me close it now :)

PR #259, opened 2020-06-03
update the way to initialize variable

kyrtaq:
it`s a really small thing but it bugs me,

in every other code example you use initialization list but in this file you didn`t do that, it would look better if you sticked to one way of initialization

### DIFF
--- a/module1/homework/fibonacci/solution/fibonacci.hpp
+++ b/module1/homework/fibonacci/solution/fibonacci.hpp
@@ -12,8 +12,8 @@ int fibonacci_iterative(int sequence) {
     if (sequence == 0 || sequence == 1)
         return sequence;
 
-    int first = 0;
-    int second = 1;
+    int first{0};
+    int second{1};
     for (size_t i = 1; i < sequence; ++i) {
         int tmp = second;
         second = first + second;


module1/homework/fibonacci/solution/fibonacci.hpp
```diff
@@ -12,8 +12,8 @@ int fibonacci_iterative(int sequence) {
     if (sequence == 0 || sequence == 1)
         return sequence;
 
-    int first = 0;
-    int second = 1;
+    int first{0};
+    int second{1};
     for (size_t i = 1; i < sequence; ++i) {
```
kontrybutor, line 17: Be consistent :) Why `i` is omitted?

kontrybutor:
> it`s a really small thing but it bugs me,
> 
> in every other code example you use initialization list but in this file you didn`t do that, it would look better if you sticked to one way of initialization

It's not initialiser list but it's called uniform initialisation.

module1/homework/fibonacci/solution/fibonacci.hpp
kyrtaq: i ommited it because in other loop in the code example it`s omitted (generateSequence), but yeah you`re right, by trying to be more consistent in the code i`m not even consistent haha how ironic

unfortunately, i made this pull request using the edu platform and now i can`t commit the changes to it ;/

PR #269, opened 2020-06-07
Homework 1

Marimeiko:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,27 @@
...
+        result = first - second;
+        std::cout << result;
+    }
+    else if (command == "divide") {
+        result = first / second;
+        std::cout << result;
+    }
+    else {
+        std::cout << "Invalid data";
+    }
```
piezol, line 25: Function 'calculate' should not have side effects (ie we should not use std::cout).

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,41 @@
...
-    // TODO: Your implementation goes here
-    return 0;
+
+    int first = 1;
+    int second = 1;
+    int result;
+
+    if (sequence <=3){
+        return 1;
+    }
```
piezol, line 12: fibonacci(3) is not equal to 1.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,41 @@
...
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
+
+    int first = 1;
+    int second = 1;
+    int result;
+
+    if (sequence <=3){
+        return 1;
+    }
```
piezol, line 30: Same thing.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,41 @@
...
+
+    for (int i = 3; i <= sequence; i++) {
+        result=first+second;
+        second=first;
+        first=result;
+        std::cout << result << std::endl;
+    }
```
piezol, line 37: It isn't really recursive, is it?

module1/homework/calculate/calculate.hpp
DamiXioR: you should just return result

module1/homework/fibonacci/fibonacci.hpp
DamiXioR: it's good practice initialize created variable etc with 0

module1/homework/fibonacci/fibonacci.hpp
DamiXioR: the same, please initialize variable

PR #270, opened 2020-06-07
Homework 1.2

Marimeiko:


module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,30 @@
...
+    else if (command == "divide") {
+        if (command==0){
+            std::cout << "Division by 0";
+        }
+        result = first / second;
+        std::cout << result;
+    }
+    else {
+        std::cout << "Invalid data";
+    }
```
piezol, line 28: I've put my comments on the wrong pull request so once again here:
Calculate function should not have any side effects (ie we should not use std::cout).

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,32 @@
...
+
+    int first = 1;
+    int second = 1;
+    int result;
+
+    if (sequence <= 3){
+        return 1;
+    }
+    
+    for (int i = 3; i <= sequence; i++) {
```
piezol, line 14: Fibonnaci (3) is not equal to 1.

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,30 @@
 #pragma once
 #include <string>
 
+int result;
```
kyrtaq, line 4: put it inside the calculate function, it`s bad practice to make global variables

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,32 @@
...
-    // TODO: Your implementation goes here
-    return 0;
+
+    int first = 1;
+    int second = 1;
+    int result;
+
+    if (sequence <= 3){
+        return 1;
+    }
```
piezol, line 12: You're not handling fib(0) at all.

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,30 @@
...
+    else if (command == "add") {
+        result = first + second;
+        std::cout << result;
+    }
+    else if (command == "subtract") {
+        result = first - second;
+        std::cout << result;
+    }
+    else if (command == "divide") {
+        if (command==0){
```
kyrtaq, line 20: you can`t compare std::string with int, did you mean to compare it with the variable called second?

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,30 @@
 #pragma once
 #include <string>
 
+int result;
+
 std::string calculate(const std::string& command, int first, int second) {
```
kyrtaq, line 6: you should return the result as a std:;string, I suggest you to use std::to_string

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,32 @@
...
-    // TODO: Your implementation goes here
-    return 0;
+
+    if(sequence == 0) {
+        return 0;
 }
+    else if (sequence < 3) {
+      return 1;
+}
+ return fibonacci_recursive(sequence-2)+fibonacci_recursive(sequence-1);
```
kuchichan, line 30: Spaces around operator. Please consider to use some autoformatter - based on Q&A lesson #1 :)

DamiXioR: it's better to check if sequence is lesser than 2 and return sequence

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,32 @@
...
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+
+    if(sequence == 0) {
+        return 0;
 }
+    else if (sequence < 3) {
+      return 1;
```
Marcinati, line 28: Or just:
if (sequence < 2) {
return sequence;
}

and other return will cover the rest (from 2).

module1/homework/fibonacci/fibonacci.hpp
WBigaj: And negative values.

module1/homework/fibonacci/fibonacci.hpp
WBigaj: Sequence < 2 is bad when int is negative.

ziobron:
Let me close this PR. If you wish to have some points please fix it, deliver and re-open this Pull Request.

PR #288, opened 2020-06-08
Homework2 - Bartek 8

g00fer:


module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,12 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+
+    for (auto number : numbers){
+        if(0 == number%2){
```
niemarcin, line 8: Would be nice to add space between if and (
As well as between number % 2.

`if (0 == number % 2) {`

Btw. you can use `(!(number % 2))`

module2/homework/addEven/addEven.hpp
```diff
@@ -2,7 +2,12 @@
 #include <vector>
 
 int addEven(const std::vector<int>& numbers) {
-    // TODO: Your implementation goes here
-    // Below return is only to make this function compile now
-    return -1;
+    int sum = 0;
+
+    for (auto number : numbers){
```
DamiXioR, line 7: I think it could be const &

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> sequence;
```
DamiXioR, line 5: you can use reserve method

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
...
+    int sum = step;
+    if(count > 0){
+        for(int i= 0; i < count; i++){
```
DamiXioR, line 8: use std::size_t is good practice

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -2,7 +2,16 @@
 #include <vector>
 #include <limits>
 
+/**
```
DamiXioR, line 5: you can remove comments

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -2,7 +2,16 @@
...
+ * Warning: 32bit architecture assumed.
+ */
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    int max = INT32_MIN;
+    
+    for (auto elem : vec){
```
DamiXioR, line 11: it could be const &

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
 #pragma once
 
+
+/**
+ * Zoptymalizowany algorytm Euklidesa
```
DamiXioR, line 5: remove comments

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
...
+    while(rhs != 0)
+    {
+		buf = rhs;
+		rhs = lhs % rhs;
+		lhs = buf;	
+	}	
+    return lhs;
 }
 
+/**
```
DamiXioR, line 27: the same

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
...
+ * From: http://www.algorytm.edu.pl/algorytmy-maturalne/nww.html
+ */ 
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+	int buf;
+	
+    if(lhs < 0){
+        lhs *= -1U;
```
DamiXioR, line 34: why "U"?

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
...
+    }
+	if(rhs < 0){
+        rhs *= -1U;
+    }
+
+    buf = NWD(lhs, rhs);
+    if(buf){
+        return lhs / buf * rhs;
+    }else{
```
DamiXioR, line 43: redundant else

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
...
-    return -1;
+    int buf;
+    
+	if(lhs < 0){
+        lhs *= -1U;
+    }
+	if(rhs < 0){
+        rhs *= -1U;
+    }
+    while(rhs != 0)
```
DamiXioR, line 18: rhs is enough

module2/homework/addEven/addEven.hpp
m-kop: no need to use & for integers

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -2,6 +2,13 @@
 #include <vector>
 
 std::vector<int> generateSequence(int count, int step) {
-    // TODO: Implement me :)
-    return {};
+    std::vector<int> sequence;
+    int sum = step;
+    if(count > 0){
```
m-kop, line 7: checking this condition before the for loop does not make any sense, compare this one with the condition in the loop

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
...
-    // TODO: Implement me :)
-    return -1;
+    int buf;
+    
+	if(lhs < 0){
+        lhs *= -1U;
+    }
+	if(rhs < 0){
+        rhs *= -1U;
+    }
```
m-kop, line 17: this is not needed, the algorithm is working fine with negative numbers as well. just return std::abs() at the very end.

module2/homework/nwdNww/nwdNww.hpp
kyrtaq, line 43: no need for that else

module2/homework/addEven/addEven.hpp
nauka-programowania-MA: For build-in type we don't need to use const&.

module2/homework/addEven/addEven.hpp
nauka-programowania-MA, line 8: Hmm, better is `(number % 2 == 0)` this is better to read :)

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA, line 5: When you know size of vector (it is count) you can reserve memory :) use `std::vector<T>::reserve()`.

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA, line 7: Check at the beginning if `count > 0` and return from function if not.

module2/homework/generateSequence/generateSequence.hpp
nauka-programowania-MA: But count is integer. Compare unsigned and signed value is worse idea.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -2,7 +2,16 @@
 #include <vector>
 #include <limits>
```
nauka-programowania-MA, line 3: remove if not use

module2/homework/maxOfVector/maxOfVector.hpp
nauka-programowania-MA: Nope, see my comment above

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
 #pragma once
 
+
+/**
+ * Zoptymalizowany algorytm Euklidesa
+ * From: http://www.algorytm.edu.pl/algorytmy-maturalne/algorytm-eulkidesa.html
+ * (I did it 20 years ago.. I promise.)
+ */
```
nauka-programowania-MA, line 8: When you copy algorithm from other side, please remove at least comment, because now I know you don't try to write it by your own :)

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
...
 int NWD(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+    int buf;
+    
+	if(lhs < 0){
+        lhs *= -1U;
+    }
```
nauka-programowania-MA, line 14: indentation

module2/homework/nwdNww/nwdNww.hpp
nauka-programowania-MA, line 17: indentation

module2/homework/nwdNww/nwdNww.hpp
nauka-programowania-MA, line 34: -1U ? You want to assign -1 as unsigned value -> which should be 0 or higher?

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,46 @@
...
 int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+	int buf;
+	
+    if(lhs < 0){
+        lhs *= -1U;
+    }
+	if(rhs < 0){
+        rhs *= -1U;
```
nauka-programowania-MA, line 37: the same, why?

module2/homework/nwdNww/nwdNww.hpp
nauka-programowania-MA, line 43: bad format, and redundant else

module2/homework/addEven/addEven.hpp
g00fer: This is a const reference to vector, not build in 'int' type. If vector consists of 1 million int's then function is effective because vector is not copied.

module2/homework/addEven/addEven.hpp
g00fer: You can read on page 3 of this book why I do like this: https://github.com/kgashok/CodeOrDie/blob/master/CoDE/docs/Expert%20C%20Programming%20-%20Deep%20C%20Secrets.pdf
(Chapter $20 Million Bug)

module2/homework/generateSequence/generateSequence.hpp
g00fer: I need to return vector. Even if count < 0 I need to return empty vector.

module2/homework/maxOfVector/maxOfVector.hpp
g00fer: I use both.

module2/homework/maxOfVector/maxOfVector.hpp
g00fer: I will leave this comment here. This function is not portable. It should be int32_t type. Then comment would not be necessary. I use INT32_MIN from limits to set the initial minimum to the smallest number. Why? Becuse if vector is null / empty I still have to return something.

module2/homework/nwdNww/nwdNww.hpp
g00fer: Meh. I did this kind of tasks in secondary school. Too old for that now :).
And honestly the comment on top of the function explains the source of the information. I put it for the sake of traceability. I see no reason to remove it, especially that the link explains what is going on below.

module2/homework/addEven/addEven.hpp
ziobron: Probably this book has some years and is dedicated to C languages. Modern compilers will warn you about assignment in if, so if it was the only reason you don't need to care about it :)
BTW it is called Yoda condition ;)

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -2,7 +2,19 @@
...
 
-int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+/**
+ * Warning: 32bit architecture assumed.
+ */
+int maxOfVector(const std::vector<int>& vec) 
+{
+    int max = INT32_MIN;
```
ziobron, line 10: `std::numeric_limits<int>::min()` instead of `INT32_MIN` is universal.

module2/homework/nwdNww/nwdNww.hpp
ziobron: Comments about what does the code do for sure should be removed. This one does not describe what does the code do but how - it's just a link to the description of the algorithm, so it's ok.

ziobron:
+4 XP addEven
+6 XP NwwNwd
+5 XP maxOfVector
+5 XP generateSequence

{
  "requirement_ids": [1631, 1633, 1634, 1635],
  "points": [4, 6, 5, 5]
}

PR #300, opened 2020-06-12
Updated info about bool and char types size

### LINKED ISSUE #272: Module 1 presentation: size of bool type - exactly 1byte / at least 1byte
In presentation about data types from module 1: https://github.com/coders-school/kurs_cpp_podstawowy/blob/master/module1/presentation_types.md

in section about bool type there is an information "_Rozmiar: co najmniej 1 bajt_", but later in this presentation, in section about char, there is an information "_Rozmiar typów logicznych i znakowych to zawsze 1 bajt._"

So the question is - does bool always have exactly 1byte or at least 1byte?

https://en.cppreference.com/w/cpp/language/types says: 
_Boolean type
bool - type, capable of holding one of the two values: true or false. The value of sizeof(bool) is implementation defined and might differ from 1._

niemarcin:
Closes #272

### DIFF
--- a/module1/presentation_types.md
+++ b/module1/presentation_types.md
@@ -38,8 +38,8 @@ ___
 
 ## Typ logiczny - `bool`
 
-* Rozmiar: co najmniej 1 bajt
-  * `sizeof(bool) == 1`
+* Rozmiar: co najmniej 1 bajt (zazwyczaj równy właśnie 1)
+  * `sizeof(bool) >= 1`
 * 2 możliwe wartości
   * `false`
   * `true`
@@ -56,7 +56,7 @@ ___
 Przedrostek `unsigned` oznacza, że typ jest bez znaku (bez liczb ujemnych), czyli od 0 do jakieś dodatniej wartości.
 <!-- .element: class="fragment fade-in" -->
 
-Rozmiar typów logicznych i znakowych to zawsze 1 bajt.
+Rozmiar typów znakowych to zawsze 1 bajt.
 <!-- .element: class="fragment fade-in" -->
 
 Rozmiary dalszych typów zależą od platformy np. 32 bity, 64 bity.


module1/presentation_types.md
```diff
@@ -38,8 +38,8 @@ ___
 
 ## Typ logiczny - `bool`
 
-* Rozmiar: co najmniej 1 bajt
-  * `sizeof(bool) == 1`
+* Rozmiar: co najmniej 1 bajt (zazwyczaj równy właśnie 1)
+  * `sizeof(bool) >= 1`
```
nauka-programowania-MA, line 42: źródło

module1/presentation_types.md
```diff
@@ -56,7 +56,7 @@ ___
 Przedrostek `unsigned` oznacza, że typ jest bez znaku (bez liczb ujemnych), czyli od 0 do jakieś dodatniej wartości.
 <!-- .element: class="fragment fade-in" -->
 
-Rozmiar typów logicznych i znakowych to zawsze 1 bajt.
+Rozmiar typów znakowych to zawsze 1 bajt.
```
nauka-programowania-MA, line 59: źródło

module1/presentation_types.md
niemarcin: https://en.cppreference.com/w/cpp/language/types

_Boolean type
bool - type, capable of holding one of the two values: true or false. The value of `sizeof(bool)` is implementation defined and might differ from 1._

Also, please take a look at #272 issue discussion

module1/presentation_types.md
niemarcin: https://en.cppreference.com/w/cpp/language/types

_Besides the minimal bit counts, the C++ Standard guarantees that
`1 == sizeof(char) <= sizeof(short) <= sizeof(int) <= sizeof(long) <= sizeof(long long)`.
Note: this allows the extreme case in which bytes are sized 64 bits, all types (including char) are 64 bits wide, and sizeof returns 1 for every type._

And bool is not always equal to 1.

module1/presentation_types.md
nauka-programowania-MA: Thank you :)

module1/presentation_types.md
nauka-programowania-MA: Yeas thank you :)

PR #302, opened 2020-06-13
added complete implementation of calculate() function

PiotrHCpp:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,19 @@
 #include <string>
 
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+    if(command == "add")
```
Kasperice, line 5: It is a good practice to use curly brackets after if/else if/else

nauka-programowania-MA:
{ 
"requirement_ids": [1429], 
"points": [5] 
}

PR #303, opened 2020-06-13
added fibonacci iterative and recursive implementation

PiotrHCpp:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,22 @@
...
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0 || sequence == 1)
+        return sequence;
+
+    int previousPreviousValue, previousValue{0}, currentValue{1};
+
+    for(std::size_t i = 1; i < sequence; ++i) {
+        previousPreviousValue = previousValue;
```
adamvm, line 10: Tricky name :)

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,22 @@
...
+    return currentValue;
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0 || sequence == 1)
+        return sequence;
+
+    return (fibonacci_recursive(sequence-2) + fibonacci_recursive(sequence-1));
```
Kasperice, line 21: If you are using whitespaces around operators, use them also here 
```suggestion
    return (fibonacci_recursive(sequence - 2) + fibonacci_recursive(sequence - 1));
```

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,22 @@
 #pragma once
 
 int fibonacci_iterative(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0 || sequence == 1)
```
nauka-programowania-MA, line 4: better
```
if (sequence < 2)
    return sequence;  // Because for 0 and 1 sequence is equal return value
```

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,22 @@
...
+        return sequence;
+
+    int secondToLastValue, previousValue{0}, currentValue{1};
+
+    for(std::size_t i = 1; i < sequence; ++i) {
```
nauka-programowania-MA, line 9: `Sequence` is `int` (so signed variable) and `i` is `size_t` (so unsigned variable) please do not compare signed and unsigned variable it's dangerous.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -1,11 +1,22 @@
...
+        previousValue = currentValue;
+        currentValue = secondToLastValue + previousValue;
+    }
+    return currentValue;
 }
 
 int fibonacci_recursive(int sequence) {
-    // TODO: Your implementation goes here
-    return 0;
+    if(sequence == 0 || sequence == 1)
```
nauka-programowania-MA, line 18: Same as above.

nauka-programowania-MA:
{ 
"requirement_ids": [1430], 
"points": [6] 
}

module1/homework/fibonacci/fibonacci.hpp
WBigaj: But then what about negative values?

module1/homework/fibonacci/fibonacci.hpp
WBigaj: The same what about negative values?

module1/homework/fibonacci/fibonacci.hpp
ziobron: Very good question. There were no test cases for checking it, so we do not require you to protect the code from it. It was one of the first tasks - that's why :) But the question is very accurate.

PR #305, opened 2020-06-21
Calculate

MaciejSalamonski:


kopia/commands.txt
```diff
@@ -0,0 +1,100 @@
...
+  304  more copy.txt
+  305  clear
+  306  ls
+  307  rm -r tmp.txt
+  308  clear
+  309  ls
+  310  history
+  311  history tail -100 > commands.txt
+  312  history tail -100n > commands.txt
+  313  history 100 > commands.txt
```
WBigaj, line 100: What is this command list?

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,25 @@
...
+    return std::to_string(first + second);
+
+    if (command == "subtract")
+    return std::to_string(first - second);
+
+    if (command == "multiply")
+    return std::to_string(first * second);
+
+    if (command == "divide")
+    return (second == 0) ? "Division by 0" : std::to_string(first / second);
```
WBigaj, line 18: I think you should use braces even for one liners. Even if you insist on not using them You should still use tab for returns.

Kasperice: As @WBigaj said, it is a very good practice to use braces and indentation, the code will be more readable when you use them. Sometimes when I am looking at some reviews is hard to say what is going on :)

module1/homework/calculate/calculate.hpp
```diff
@@ -1,7 +1,25 @@
...
+
+    else
+    return "Invalid data";
+    
     return "";
```
Kasperice, line 23: Is this return necessary?

ziobron:
{ "requirement_ids": [1429] }
+5 XP granted

PR #306, opened 2020-06-21
Homework2

Papierkowy:
I decided to put it in, but I couldn't run the tests for Nwd. When I get help it will solve this problem


Papierkowy:
I don't add one commit in NWD and NWW.

PR #307, opened 2020-06-21
Fibonacci

MaciejSalamonski:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,26 @@
 
 int fibonacci_iterative(int sequence) {
     // TODO: Your implementation goes here
-    return 0;
+
+    if ( (sequence == 0) or (sequence == 1)) {
+        return sequence;
+    }
```
WBigaj, line 8: You are not handling negative values here.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,26 @@
...
+
+    return sequenceNumber;
 }
 
 int fibonacci_recursive(int sequence) {
     // TODO: Your implementation goes here
-    return 0;
+    
+    return ((sequence == 0) or (sequence == 1) ? sequence : fibonacci_recursive(sequence - 2) + fibonacci_recursive(sequence - 1));
```
WBigaj, line 26: It will crash if you would call it for negative value.

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,26 @@
 
 int fibonacci_iterative(int sequence) {
     // TODO: Your implementation goes here
```
Marcinati, line 4: You should remove this comment.

module1/homework/fibonacci/fibonacci.hpp
MaciejSalamonski: Thanks for the advice ;].

Marcinati: Better option: 
``` cpp
if (sequence < 2) {
    return sequence; 
}

File: module1/homework/fibonacci/fibonacci.hpp [hunk 4]
```diff
@@ -2,10 +2,26 @@
...
+        sequenceNumber = beforePreviousNumber + previousNumber;
+        beforePreviousNumber = previousNumber;
+        previousNumber = sequenceNumber;
+    }
+
+    return sequenceNumber;
 }
 
 int fibonacci_recursive(int sequence) {
     // TODO: Your implementation goes here
```
Marcinati, line 24: You can delete that line.

module1/homework/fibonacci/fibonacci.hpp
Marcinati, line 26: Almost good (besides cases for negative values), but long lines are harder to read.

module1/homework/fibonacci/fibonacci.hpp
WBigaj: Problem with this aproach is that for negative values(invalid data) you have ton of possible outcomes instead of one single specified error code(like -1).

ziobron:
{ "requirement_ids": [1430] }
+6 XP granted.

PR #308, opened 2020-06-23
Homework2

MaciejSalamonski:


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,20 @@
...
+    {
+        for(std::size_t i = 0; i != count; i++)
+        {
+            result.push_back( step * (i+1) );
+        }
+
+    return result;
+    }
+
+    else
```
kuchichan, line 18: You can make `if (count <= 0) return 0` and later you will not need this `else`

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,30 @@
 #pragma once
+#include<cmath>
+
+int NWD(int lhs, int rhs){
+    
+   int reszta = {};
+
+   while( rhs != 0)
+   {
+       reszta = lhs % rhs;
```
kuchichan, line 10: English pls :)

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,30 @@
...
-    return -1;
 }
 
-int NWW(int lhs, int rhs) {
-    // TODO: Implement me :)
-    return -1;
+int NWW(int lhs, int rhs){
+    
+    if(NWD(lhs, rhs) == 0)
+    {
```
kuchichan, line 21: Please introduce some auto formatter :)

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,30 @@
...
+    return 0;
+    }
+
+    else
+    {
+    return abs((lhs*rhs) / NWD(lhs, rhs));
```
Vicx95, line 27: indentation

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,20 @@
...
-    // TODO: Implement me :)
+std::vector<int> generateSequence(int count, int step) 
+{
+    std::vector<int> result;
+
+    if(count > 0)
+    {
+        for(std::size_t i = 0; i != count; i++)
+        {
+            result.push_back( step * (i+1) );
```
WBigaj, line 12: Instead of multiplying step you should consider increasing by adding count every iteration. It's faster for larger size data.

module2/homework/generateSequence/generateSequence.hpp
WBigaj: Agree

module2/homework/generateSequence/generateSequence.hpp
MaciejSalamonski: Thanks for that one. It should be:

result.push_back(step);
step += step;

Did you mean that?

WBigaj: No,because then for step=100 you would have 100,200,400,800 etc.

I meant something like this:
int sum= 0;

sum += step;
result.push_back(sum);

MaciejSalamonski: In the first reply I wrote exactly the same as you, but i think there is no point in using an additional variable. Appart from using
Int sum = 0;, you can write sth like this:

result.push_back(step);
step += step;

Lets suppose that variable step is 3

Iteration 1: 
step = 3 
result.push_back(step);
step += step;  - > step = 6

Iteration 2:

step = 6
result.push_back(step);
step += step;  - > step = 9

e.t.c.

Result of that is the same but you're saving at least 2 bytes of memory. What do you think about that?

WBigaj: In second iteration step = 6,so doing step += step should give 12 not 9.

MaciejSalamonski: Opps i made mistake. Thanks for your replies

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -1,8 +1,11 @@
 #pragma once
 #include <vector>
 #include <limits>
+#include <algorithm>
```
anraf1001, line 4: You should keep your includes in alphabetical order

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -1,8 +1,11 @@
...
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    
+    int result = *std::max_element(vec.begin(), vec.end());
+
+    return result;
```
anraf1001, line 10:
```suggestion
    return *std::max_element(vec.begin(), vec.end());
```

ziobron:
{ "requirement_ids": [1631, 1633, 1634, 1635] }
4 XP za Podstawy#2 – addEven (1631)
6 XP za Podstawy#2 – NWD i NWW (1633)
5 XP za Podstawy#2 – maxOfVector (1634)
5 XP za Podstawy#2 – generateSequence (1635)

PR #317, opened 2020-07-24
Homework2

Ptysiek:
Quite a delay

### DIFF
--- a/module2/homework/generateSequence/generateSequence.hpp
+++ b/module2/homework/generateSequence/generateSequence.hpp
@@ -1,7 +1,18 @@
 #pragma once
 #include <vector>
+#include <iostream>
 
-std::vector<int> generateSequence(int count, int step) {
+std::vector<int> generateSequence(int count, int step) {    
+    if (count < 1) {
+        return std::vector<int> {};
+    }
+    
+    std::vector<int> result(count, step);
+
+    for (size_t i = 1; i < count; ++i) {
+        result[i] += step * i;
+    }
+
     // TODO: Implement me :)
-    return {};
+    return result;
 }
--- a/module2/homework/maxOfVector/maxOfVector.hpp
+++ b/module2/homework/maxOfVector/maxOfVector.hpp
@@ -3,6 +3,21 @@
 #include <limits>
 
 int maxOfVector(const std::vector<int>& vec) {
-    // TODO: Implement me :)
-    return {};
+    
+    int result;
+
+    if (vec.size() > 0) {
+        result = vec[0];
+    }
+    else {
+        return -1;
+    }
+
+    for (const auto& value : vec) {
+        if (value > result) {
+            result = value;
+        }
+    }
+
+    return result;
 }
--- a/module2/homework/nwdNww/nwdNww.hpp
+++ b/module2/homework/nwdNww/nwdNww.hpp
@@ -23,15 +23,29 @@ std::map<int, int> getPrimals(int value) {
     return result;
 }
 
+void ShowPrimals (const std::map<int,int>& map) {
+    for (auto const& [key, value] : map) {
+        std::cout << "[" << key << ", " << value << "] \n";
+    }
+    std::cout << "\n";
+}
+
 
 int NWW(int lhs, int rhs) {
+    if (lhs == 0 || rhs == 0) {
+        return 0;
+    }
+
     std::map<int,int> lprim { getPrimals(lhs) };
     std::map<int,int> rprim { getPrimals(rhs) };
     int result = 1;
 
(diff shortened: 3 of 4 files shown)


module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,18 @@
 #pragma once
 #include <vector>
+#include <iostream>
```
WBigaj, line 3: You should get a habit to sort includes alphabetically. Also new line between include and pragma would be nice.

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,18 @@
...
+
+std::vector<int> generateSequence(int count, int step) {    
+    if (count < 1) {
+        return std::vector<int> {};
+    }
+    
+    std::vector<int> result(count, step);
```
WBigaj, line 10: So here you are creating vector with step value placed count times. For what? Isn't better to just create count-sized vector to reserve memory?

module2/homework/generateSequence/generateSequence.hpp
```diff
@@ -1,7 +1,18 @@
...
+
+    for (size_t i = 1; i < count; ++i) {
+        result[i] += step * i;
+    }
```
WBigaj, line 14: You should consider additive aproach instead of multiplying step every single time. Even you need extra variable to store incrementing sum it will be faster for larger data.

module2/homework/maxOfVector/maxOfVector.hpp
```diff
@@ -3,6 +3,21 @@
...
-    return {};
+    
+    int result;
+
+    if (vec.size() > 0) {
+        result = vec[0];
+    }
+    else {
+        return -1;
+    }
```
WBigaj, line 14: Keep in mind that when you return you espace function.
You can utilise this to simplify code like this:
if(vec.size() <= 0){
    return -1;
}
result = vec[0];

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,83 @@
 #pragma once
+#include <map>
```
WBigaj, line 2: Here new line also would be nice.

module2/homework/nwdNww/nwdNww.hpp
```diff
@@ -1,11 +1,83 @@
...
+    if (lhs == 0 || rhs == 0) {
+        return 0;
+    }
+
+    std::map<int,int> lprim { getPrimals(lhs) };
+    std::map<int,int> rprim { getPrimals(rhs) };
+    int result = 1;
+
+    // ShowPrimals(lprim);    
+    // ShowPrimals(rprim);
```
WBigaj, line 44: Do not leave commented code in the program.

module2/homework/generateSequence/generateSequence.hpp
Ptysiek: When vector is created with a single argument `count`, you not only reserve the amount of space specified in the argument, but also create the given number of elements. For the `int` type, those elements are initialized with value 0 (zero).

I was thinking "what's the difference to initialize with a constant value of zero and any other value?"
I checked it, and it makes no difference for cpu.

Also, found this fun fact:
Screenshot_2020-07-27_14-38-20

What I did not expect. It turns out my solution is actually faster. 
BUT, IT IS ONLY A FUN FACT.
Because these results are due to the optimization of the clang 9.0 compiler.
On the other compilars, (ex. GCC 10.1), both methods take on average the same time.
Same thing when optimizations are turned off. `benchmark::DoNotOptimize(testVector);`

Ptysiek: > So here you are creating vector with step value placed count times. For what? Isn't better to just create count-sized vector to reserve memory?

The first element of the vector is already initialized with appropriate value.
So it allows me to skip it, and start the `for` loop at index 1.
Not necessary, probably less readable. Wanted to try out a new thing I learned.

WBigaj: Nice benchmark. I have done some by myself and found even faster way.
https://quick-bench.com/q/yK2UadqrqHcHiV-6r1A5iFhBOOs

ziobron:
4 XP za Podstawy#2 – addEven (1631)
6 XP za Podstawy#2 – NWD i NWW (1633)
5 XP za Podstawy#2 – maxOfVector (1634)
5 XP za Podstawy#2 – generateSequence (1635)

{ "requirement_ids": [1631, 1633, 1634, 1635] }

PR #320, opened 2020-08-02
fibonacci work

EmKalin:


module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,40 @@
...
+		
+			c = 1;
+		} else {
+		
+			a=b;
+			b=c;
+			c=a+b;
+		}
+		
+
```
WBigaj, line 21: Firsly what are those whole empty spaces?
Secondly can't you do it while using only two variables?

module1/homework/fibonacci/fibonacci.hpp
```diff
@@ -2,10 +2,40 @@
...
 
 int fibonacci_recursive(int sequence) {
     // TODO: Your implementation goes here
-    return 0;
+
+    	int solution = 0;
+
+	if (sequence < 1){
+	
+		solution = 0;
```
WBigaj, line 35: I know it's not defined in task how negative values should be handled,but I think that returning error code like -1 is better than just returning 0.

PR #326, opened 2020-08-12
Homework 1.1 - Calculate

Marimeiko:


module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,23 @@
...
+    std::string result;
+
+    if (command == "add") {
+        result = std::to_string(first + second);
+    } else if (command == "subtract") {
+        result = std::to_string(first - second);
+    } else if (command == "multiply") {
+        result = std::to_string(first * second);
+    } else if (command == "divide") {
+        if (second = 0) {
```
ziobron, line 14:
```suggestion
        if (second == 0) {
```
Przypisanie masz zamiast porównania. W if do second przypisujesz wartość 0 zamiast porównać czy second jest równe 0 i jest problem :)

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,23 @@
 #include <string>
 
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+    std::string result;
+
+    if (command == "add") {
+        result = std::to_string(first + second);
```
WBigaj, line 8: Consider simplyfing program by just returning values like std::to_string(first + second) instead of assigning them to later returned value(which should not even by changed). Then program does not need to go through rest of the code then.
Moreover you can shorten program by removing elses.

module1/homework/calculate/calculate.hpp
Marimeiko: I have done some changes in the newest commit, please take a look

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,27 @@
 #include <string>
 
 std::string calculate(const std::string& command, int first, int second) {
-    // TODO: Implement your solution here and return proper value
-    return "";
+    std::string result;
```
WBigaj, line 5: You are not using this variable,so it's redundant. This should generate warning and It wouldn't compile if you would use all gcc warning flags: -Wall -Wextra -pedantic -Wconversion.

module1/homework/calculate/calculate.hpp
```diff
@@ -2,6 +2,27 @@
...
+
+    if (command == "divide") {
+        if (second == 0) {
+            return "Division by 0";
+        } else {
+            return std::to_string(first / second);
+        }
+    }
+
+    return "Invalid command";
```
ziobron, line 27: There are some problems in unit tests:
1: Expected equality of these values:
1:   calculate("quit", 10, 20)
1:     Which is: "Invalid command"
1:   "Invalid data"

They are expecting "Invalid data" string instead of "Invalid command".

ziobron (approved):

ziobron:
{ "requirement_ids": [1429], "points": [5] }

PR #335, opened 2020-09-22
Update presentation_list.md

majkel84:
little correction list instead map for begin() and end()


WBigaj (approved):
I have been reading this presentation lately and have not noticed it xD Good catch :)

ziobron (approved):

